PR #28, opened 2026-03-22
CA-547: feat: Implement the core user management module including the…

SimonDerejeW:
# PR Review: feat/migrate-user-feature-to-schema → feat/migrate/cost-files

### 1. PR SUMMARY
This PR migrates the core user features to a schema-first structure under `supabase/schemas/core/users/`. It introduces the `users` table, a `user_profiles` public view for safe display of user data, performance indexes, a securely scoped function to check email availability (`check_email_exists`), row-level security policies granting users full access to their own profiles, and comprehensive documentation in `README.md`.

### 2. REVIEW SUMMARY TABLE

| Missing `updated_at` Trigger | The `users` table has an `updated_at` column, but there is no corresponding database trigger to ensure it automatically updates on row modifications. (Missing `04_triggers.sql`) | ✅ |  |

### DIFF
--- a/supabase/schemas/core/users/02_indexes.sql
+++ b/supabase/schemas/core/users/02_indexes.sql
@@ -1,0 +1,16 @@
+-- Users Indexes
+-- Performance indexes for common query patterns
+
+CREATE INDEX "users_created_at_idx" ON "public"."users" USING "btree" ("created_at");
+
+
+CREATE INDEX "users_professional_role_idx" ON "public"."users" USING "btree" ("professional_role");
+
+
+CREATE INDEX "users_user_status_idx" ON "public"."users" USING "btree" ("user_status");
+
+
+CREATE INDEX "idx_users_credential_id" ON "public"."users" USING "btree" ("credential_id");
+
+
+CREATE INDEX "idx_users_id_credential" ON "public"."users" USING "btree" ("id", "credential_id");
--- a/supabase/schemas/core/users/02_indexes.sql
+++ b/supabase/schemas/core/users/02_indexes.sql
@@ -1,16 +1,10 @@
 -- Users Indexes
 -- Performance indexes for common query patterns
 
-CREATE INDEX "users_created_at_idx" ON "public"."users" USING "btree" ("created_at");
-
-
 CREATE INDEX "users_professional_role_idx" ON "public"."users" USING "btree" ("professional_role");
 
 
 CREATE INDEX "users_user_status_idx" ON "public"."users" USING "btree" ("user_status");
 
 
-CREATE INDEX "idx_users_credential_id" ON "public"."users" USING "btree" ("credential_id");
-
-
-CREATE INDEX "idx_users_id_credential" ON "public"."users" USING "btree" ("id", "credential_id");
+CREATE INDEX "users_created_at_idx" ON "public"."users" USING "btree" ("created_at");
--- a/supabase/schemas/_shared/01_functions.sql
+++ b/supabase/schemas/_shared/01_functions.sql
@@ -1,0 +1,15 @@
+-- Shared General Functions
+
+-- Function to automatically update the updated_at timestamp when a row is modified
+CREATE OR REPLACE FUNCTION "public"."set_current_timestamp_updated_at"()
+    RETURNS TRIGGER
+    LANGUAGE "plpgsql"
+    AS $$
+BEGIN
+  NEW.updated_at = "now"();
+  RETURN NEW;
+END;
+$$;
+
+ALTER FUNCTION "public"."set_current_timestamp_updated_at"() OWNER TO "postgres";
+COMMENT ON FUNCTION "public"."set_current_timestamp_updated_at"() IS 'Automatically updates updated_at column to current timestamp on row modification.';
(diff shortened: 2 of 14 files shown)


supabase/schemas/core/users/02_indexes.sql
```diff
@@ -0,0 +1,16 @@
...
+CREATE INDEX "users_professional_role_idx" ON "public"."users" USING "btree" ("professional_role");
+
+
+CREATE INDEX "users_user_status_idx" ON "public"."users" USING "btree" ("user_status");
+
+
+CREATE INDEX "idx_users_credential_id" ON "public"."users" USING "btree" ("credential_id");
+
+
+CREATE INDEX "idx_users_id_credential" ON "public"."users" USING "btree" ("id", "credential_id");
```
ripplearcgit: 🔴 **Redundant composite index + naming inconsistency.**

`idx_users_id_credential (id, credential_id)` provides no query benefit — `id` is the primary key, so filtering by `id AND credential_id` together is never more efficient than filtering by `id` alone. Drop it.

Also: `idx_users_credential_id` and `idx_users_id_credential` use the `idx_users_*` prefix while the other three use `users_*_idx`. Pick one convention. Suggested clean set:

```sql
CREATE INDEX "users_credential_id_idx"     ON "public"."users" USING "btree" ("credential_id");
CREATE INDEX "users_professional_role_idx" ON "public"."users" USING "btree" ("professional_role");
CREATE INDEX "users_user_status_idx"       ON "public"."users" USING "btree" ("user_status");
CREATE INDEX "users_created_at_idx"        ON "public"."users" USING "btree" ("created_at");
-- idx_users_id_credential dropped: redundant with PK
```

ripplearcgit (changes requested):
Four items to address — two blocking, two non-blocking, plus a nit.

**🔴 `02_indexes.sql` — redundant composite index + naming inconsistency**
`idx_users_id_credential (id, credential_id)` provides no benefit: `id` is already the primary key with its own unique index, so no realistic query would filter by `id AND credential_id` together. Additionally, three indexes use `users_*_idx` naming while two use `idx_users_*` — these should be consistent. See inline comment.

**🔴 `04_views.sql` + `05_rls.sql` — duplicate `04_` prefix**
Both `04_triggers.sql` and `04_views.sql` share the same numeric prefix. If files are ever applied in alphabetical/glob order (the obvious intent of the numbering scheme), the load order between them is ambiguous. `04_views.sql` should be renumbered to `05_views.sql` and `05_rls.sql` to `06_rls.sql`. See inline comment.

**🟡 `04_views.sql` — `credential_id` leaks through the "public" view**
The view is designed to expose a safe public subset of user data, but `credential_id` is the direct foreign key to `auth.users`. Exposing it means any caller with SELECT on `user_profiles` can correlate app users to auth identities — the opposite of the view's stated purpose. See inline comment.

**🟡 `05_rls.sql` — single policy permits unguarded INSERT**
A combined USING + WITH CHECK policy on all operations means any authenticated user can INSERT a new row for themselves, potentially creating duplicate profile rows. User profile creation should be gated to a trusted function (e.g., an `AFTER INSERT` trigger on `auth.users`); the RLS policy should cover SELECT/UPDATE only, or at minimum split into explicit per-operation policies. See inline comment.

**🟢 `_shared/01_functions.sql` — missing `COMMENT ON FUNCTION`**
Every other function in the codebase has a `COMMENT ON FUNCTION` statement. See inline comment.

ripplearcgit (approved):
## Re-Review (R1 → R2)

**Prior review SHA:** `b3142af` | **Current HEAD:** `a8639c0` — new commits reviewed.

All 5 prior findings addressed. ✅

---

### Prior findings — resolution status

| Finding | Status |
|---|---|
| Redundant composite index `idx_users_id_credential` + naming inconsistency | ✅ Resolved — index removed, all names now `users_*_idx` |
| Duplicate `04_` prefix on views/RLS files | ✅ Resolved — now `04_triggers`, `05_views`, `06_rls` |
| `credential_id` leaking through `user_profiles` view | ✅ Resolved — view exposes only safe fields |
| Unguarded INSERT via combined RLS policy | ✅ Resolved — split into SELECT/UPDATE policies; INSERT excluded with comment |
| Missing `COMMENT ON FUNCTION` in `_shared/01_functions.sql` | ✅ Resolved — comment added |

---

### New findings (non-blocking)

🟡 **`users/02_indexes.sql` — `users_credential_id_idx` is redundant**
The `UNIQUE` constraint on `credential_id` (`users_credential_id_key`) already creates an implicit B-tree index in Postgres. The explicit `CREATE INDEX "users_credential_id_idx"` duplicates it with no benefit and extra write overhead. Suggest dropping the explicit index.

🟡 **`05_views.sql` + `06_rls.sql` — `user_profiles` view can't serve its stated cross-user purpose with current RLS**
The README says to use `user_profiles` "for viewing other users (e.g., in project teams)" but the view uses `SECURITY INVOKER` (default), so the `users_select_own` RLS policy (`auth.uid() = credential_id`) applies — a user can only ever see their own row through the view. If cross-user reads are deferred, the README should say so explicitly rather than implying the view already enables that.

🟡 **`cost_files/03_rls.sql` — deferred RLS has no YouTrack ticket link**
Per the convention enforced as a blocker in this PR's companion PR #29, deferred RLS must carry a `-- TODO: [CA-XXX] https://...` ticket link before merge. The comment explains intent but doesn't link a ticket.

---

✅ **APPROVE** — no blockers remain. The three nits above can be addressed before or after merge at your discretion.

PR #30, opened 2026-03-27
feat: add global search backend migrations

Bazabizi:
## Summary

This PR introduces the complete backend data layer for the Global Search feature. It adds a new `search_history` table with five Row Level Security policies (four for personal CRUD ownership and one for cross-user teammate visibility on shared projects), two public RPCs (`global_search` and `get_search_suggestions`) with `SECURITY INVOKER` to enforce RLS, two `BEFORE UPDATE` triggers to atomically manage `search_count` and `updated_at`, a full suite of supporting indexes, and a comprehensive module README. The implementation ships as both migration files (applied to the live database) and canonical schema source files (serving as the single source of truth), and includes rollback scripts for all database objects.

---

### Impact Stats

| Category               | Value                        |
|------------------------|------------------------------|
| **Files Added**        | 9                            |
| **Files Modified**     | 0                            |
| **Files Deleted**      | 0                            |
| **Migration SQL Lines**| 406 (26 + 57 + 270 + 53)    |
| **Schema SQL Lines**   | 397 (29 + 49 + 292 + 27)    |
| **Documentation Lines**| 135 (README.md)              |
| **Total Insertions**   | 938                          |
| **PR Size (Production Code)** | **L** (200+ lines)  |
| **New Tables**         | 1 (`search_history`)         |
| **New RPCs**           | 2 (`global_search`, `get_search_suggestions`) |
| **New Triggers**       | 2                            |
| **New RLS Policies**   | 5                            |
| **Rollback Scripts**   | Present in all migration files |

> **Note on PR Size:** This PR is classified **L**. However, all 9 files belong to a single, cohesive feature unit (schema definition + migration + documentation). Breaking it further would create artificial coupling between migration steps and is not recommended in this case. The size is justified.

---

## Review Summary
|---|---|---|---|
| Schema File Naming Convention Deviation | The project convention (per Rule 7) is: `01_table.sql`, `02_indexes.sql`, `03_functions.sql`, `04_triggers.sql`, `05_rls.sql`. This module uses: `01_table.sql` (includes indexes inline), `02_rls.sql`, `03_functions.sql`, `04_triggers.sql` — deviating from the standard ordering. Indexes are embedded in `01_table.sql` rather than extracted to a dedicated `02_indexes.sql`, and RLS is at position `02` rather than `05`. Author should verify whether this deviation is intentional for this module type or whether a separate `02_indexes.sql` and a renaming of `02_rls.sql` → `05_rls.sql` is needed for consistency with other schema modules. | Fixed | |
| Boolean Index Low Cardinality | `search_history_has_results_idx` is an index on the `has_results` boolean column. PostgreSQL's query planner typically ignores low-cardinality indexes (especially boolean columns) in favour of sequential scans when a significant fraction of rows share a value. This index may provide little benefit at scale. Author should evaluate whether a partial index (e.g., `WHERE has_results = true`) would be more effective for the suggestion query path. | Fixed | search_history_has_results_idx changed to a partial index WHERE has_results = true in both the migration file and schema 02_indexes.sql. |


ripplearcgit (changes requested):
## Code Review — Global Search Backend

Three blocking issues below. Non-blocking findings posted separately.

---

### 🔴 B1 — `credential_id` leaks through `global_search` members result

`global_search` exposes `u.credential_id` in the members SELECT list. `credential_id` is the Supabase `auth.uid()` — an internal auth identifier that must never surface in public-facing API responses. Any client calling this RPC receives the `credential_id` for every member in the result set.

**File:** `supabase/migrations/20260324000003_global_search_rpc.sql` (and `supabase/schemas/global_search/search_history/03_functions.sql`)

```sql
-- Remove credential_id from the members SELECT
SELECT DISTINCT ON (u.id)
  u.id,
  -- u.credential_id,  ← remove this line
  u.first_name,
  u.last_name,
  u.professional_role,
  u.profile_photo_url
FROM project_members pm
JOIN users u ON pm.user_id = u.id
...
```

---

### 🔴 B2 — Deferred work (`filter_by_tag`, orphan cleanup, `pg_trgm`) lacks TODO + YouTrack links

Three pieces of deferred work are documented in comments or the README with no YouTrack ticket attached. Per project convention, any work intentionally deferred to a future PR **must** have a TODO comment with a YouTrack issue link in the SQL file or README.

**1. `filter_by_tag` no-op** — in both migration and schema `03_functions.sql`:
```sql
-- TODO: [CA-XXX] Wire up filter_by_tag once project-tag schema exists.
-- https://ripplearc.youtrack.cloud/issue/CA-XXX
```

**2. Orphan cleanup** — in `README.md`:
```markdown
> **TODO ([CA-XXX]):** Implement periodic cleanup job to purge search_history rows
> where user_id no longer exists in auth.users. [CA-XXX](https://ripplearc.youtrack.cloud/issue/CA-XXX)
```

**3. `pg_trgm` / full-text search follow-up** — noted in README performance section:
```markdown
> **TODO ([CA-XXX]):** Add pg_trgm GIN indexes or full-text search to replace leading-wildcard LIKE.
> [CA-XXX](https://ripplearc.youtrack.cloud/issue/CA-XXX)
```

YouTrack issues must exist before merge.

---

### 🔴 B3 — Migration files hand-authored rather than generated from schema

The `supabase/schemas/` files are described as the "single source of truth," but the migration files in `supabase/migrations/` were written manually and kept in sync by hand. This is evident from structural divergences between the two layers:

- Schema files use `"public"."search_history"` (explicit schema prefix); migration files omit it
- Schema `01_table.sql` uses `CREATE TABLE IF NOT EXISTS`; migration uses `CREATE TABLE`
- Schema `04_triggers.sql` uses `CREATE OR REPLACE TRIGGER`; migration uses `CREATE TRIGGER`

The correct workflow is:
1. Developer edits `supabase/schemas/` files only
2. Runs `supabase db diff -f <migration_name>` to auto-generate the migration
3. Reviews and commits both

Hand-authoring both layers creates two places to make mistakes and guarantees drift over time. The migration in this PR should be regenerated via `supabase db diff` from the schema files. The team should also establish this as the enforced workflow going forward (CI check or documented rule) so subsequent PRs don't repeat the same pattern.

ripplearcgit (commented):
## Non-blocking findings (Supplementary)

---

### 🟡 N1 — Per-entity `LIMIT` breaks pagination semantics

`global_search` applies `LIMIT`/`OFFSET` independently to each of the three entity queries. With `limit = 20`, the function can return up to **60 rows** total. The README acknowledges this ("Up to `3 × limit` rows may be returned") but it creates a broken pagination API — `offset` advances all three cursors simultaneously, which will skip or duplicate results inconsistently across entity types.

Consider either accepting per-entity limits, or returning total counts per entity alongside results so clients can paginate independently.

---

### 🟡 N2 — Leading wildcard `LIKE` prevents index use

```sql
v_search := '%' || lower(query) || '%';
```

A leading `%` defeats B-tree indexes — every text match is a sequential scan. This is acceptable for an initial implementation but will not scale. The README notes this already; the TODO + YouTrack link for `pg_trgm` GIN indexes is covered in B2 above.

---

### 🟡 N3 — `scope` parameter is unvalidated

An unknown value (e.g. `'members'` instead of `'member'`) silently returns an empty result with no error. Add an explicit guard:

```sql
IF scope IS NOT NULL AND scope NOT IN ('dashboard', 'estimation', 'member') THEN
  RAISE EXCEPTION 'Invalid scope: %', scope USING ERRCODE = '22023';
END IF;
```

---

### 🟡 N4 — Teammate SELECT policy missing `project_id IS NOT NULL` short-circuit

The two SELECT policies on `search_history` are OR-combined by Postgres. The teammate policy runs its `project_members` subquery on every SELECT, including personal history reads where `project_id IS NULL` and can never match. Add a short-circuit to avoid the join on null rows:

```sql
CREATE POLICY "search_history_teammate_select_policy" ON "search_history"
  FOR SELECT
  USING (
    project_id IS NOT NULL   -- ← add this guard
    AND project_id IN (
      SELECT pm.project_id
      FROM project_members pm
      JOIN users u ON pm.user_id = u.id
      WHERE u.credential_id = auth.uid()
        AND pm.membership_status = 'joined'
    )
  );
```

ripplearcgit (approved):
## Conditional LGTM ✅

All three blocking findings from the previous round are resolved:

- **B1** — `credential_id` removed from the members SELECT in both migration and schema.
- **B2** — All deferred work annotated with TODO + YouTrack links (CA-596 `filter_by_tag`, CA-597 orphan cleanup, CA-598 `pg_trgm`).
- **B3** — Hand-authored migrations replaced with a single `supabase db diff`-generated file (`20260401085103_32_search_tables.sql`). N3 (scope validation) and N4 (`project_id IS NOT NULL` guard) also addressed.

Two non-blocking items remain open:

### 🟡 N1 — Per-entity LIMIT pagination semantics (carry-forward, unaddressed)

`global_search` applies `LIMIT`/`OFFSET` independently to each of the three entity queries, meaning up to `3 × limit` rows can be returned and `offset` advances all three cursors simultaneously. A client paginating through projects will also skip estimation and member results at the same rate. Not a blocker for this PR, but should be tracked.

### 🟡 README Migration Files table is stale

The "Migration Files" section still lists the four original hand-authored files. The actual migration is now a single consolidated file `20260401085103_32_search_tables.sql`. Please update before merge.

PR #31, opened 2026-04-12
feat: migrate cost_estimates RLS policies and triggers to use JWT 

SimonDerejeW:
# PR Review: feat/migrate-cost-estimate-policies-to-jwt-auth → master

**PR Branch:** feat/migrate-cost-estimate-policies-to-jwt-auth (with changes)
**Base Branch:** master (target)
**Date:** Sun Apr 12 16:32:09 EAT 2026
**Reviewer:** Claude Code
**Review Date:** 2026-04-12

---

## 📋 PR SUMMARY

This PR migrates the `cost_estimates` table's RLS policies and trigger functions from database-lookup-based permission checks to JWT-claims-based permission checks. This is a performance optimization that eliminates expensive per-row database joins by reading permissions directly from JWT tokens in memory. The PR also fixes critical bugs in the test suite where JWT claims were not properly structured, and corrects a bug in the trigger function where `locked_by_user_id` was not being set correctly.

### DIFF
--- a/supabase/migrations/20260313081301_auth_hook_project_permission_claims.sql
+++ b/supabase/migrations/20260313081301_auth_hook_project_permission_claims.sql
@@ -8,8 +8,14 @@ AS $$
 DECLARE
   claims jsonb;
   projects_claims jsonb;
+  user_internal_id uuid;
 BEGIN
   claims := event->'claims';
+
+  -- Get the internal user ID (users.id) from credential_id
+  SELECT u.id INTO user_internal_id
+  FROM public.users u
+  WHERE u.credential_id = (event->>'user_id')::uuid;
 
   SELECT COALESCE(
     jsonb_object_agg(project_permissions.project_id, project_permissions.permissions),
@@ -29,12 +35,15 @@ BEGIN
     GROUP BY pm.project_id
   ) AS project_permissions;
 
-  -- Ensure app_metadata exists, then merge in projects
+  -- Ensure app_metadata exists, then merge in projects and internal_user_id
   -- This handles cases where claims don't yet have app_metadata (e.g., first login)
   claims := jsonb_set(
     claims,
     '{app_metadata}',
-    COALESCE(claims->'app_metadata', '{}'::jsonb) || jsonb_build_object('projects', projects_claims),
+    COALESCE(claims->'app_metadata', '{}'::jsonb) || jsonb_build_object(
+      'projects', projects_claims,
+      'internal_user_id', user_internal_id
+    ),
     true
   );
 
--- a/supabase/migrations/20260313081637_migrate_cost_estimates_rls_to_jwt.sql
+++ b/supabase/migrations/20260313081637_migrate_cost_estimates_rls_to_jwt.sql
@@ -34,6 +34,9 @@ CREATE POLICY "cost_estimates_update_policy" ON "cost_estimates"
   TO authenticated
   USING (
     jwt_has_project_permission(project_id, 'edit_cost_estimation')
+  )
+  WITH CHECK (
+    jwt_has_project_permission(project_id, 'edit_cost_estimation')
   );
 
 -- Policy for DELETE operations (removing cost estimates)
--- a/supabase/migrations/20260313081637_migrate_cost_estimates_rls_to_jwt.sql
+++ b/supabase/migrations/20260313081637_migrate_cost_estimates_rls_to_jwt.sql
@@ -1,0 +1,46 @@
+-- Migrate cost_estimates RLS policies to use JWT claims instead of database lookups
+-- This improves performance by reading permissions from JWT tokens rather than querying the database
+
+-- Drop existing RLS policies
+DROP POLICY IF EXISTS "cost_estimates_select_policy" ON "cost_estimates";
+DROP POLICY IF EXISTS "cost_estimates_insert_policy" ON "cost_estimates";
+DROP POLICY IF EXISTS "cost_estimates_update_policy" ON "cost_estimates";
+DROP POLICY IF EXISTS "cost_estimates_delete_policy" ON "cost_estimates";
+
+-- Create new JWT-based policies
+
+-- Policy for SELECT operations (viewing cost estimates)
+-- Users can view cost estimates if they have the 'get_cost_estimations' permission in their JWT
+CREATE POLICY "cost_estimates_select_policy" ON "cost_estimates"
+  FOR SELECT
+  TO authenticated
+  USING (
+    jwt_has_project_permission(project_id, 'get_cost_estimations')
+  );
+
+-- Policy for INSERT operations (creating cost estimates)
+-- Users can create cost estimates if they have the 'add_cost_estimation' permission in their JWT
+CREATE POLICY "cost_estimates_insert_policy" ON "cost_estimates"
+  FOR INSERT
+  TO authenticated
+  WITH CHECK (
+    jwt_has_project_permission(project_id, 'add_cost_estimation')
+  );
+
+-- Policy for UPDATE operations (editing cost estimates)
+-- Users can update cost estimates if they have the 'edit_cost_estimation' permission in their JWT
+CREATE POLICY "cost_estimates_update_policy" ON "cost_estimates"
+  FOR UPDATE
+  TO authenticated
+  USING (
+    jwt_has_project_permission(project_id, 'edit_cost_estimation')
+  );
+
+-- Policy for DELETE operations (removing cost estimates)
+-- Users can delete cost estimates if they have the 'delete_cost_estimation' permission in their JWT
+CREATE POLICY "cost_estimates_delete_policy" ON "cost_estimates"
+  FOR DELETE
+  TO authenticated
+  USING (
+    jwt_has_project_permission(project_id, 'delete_cost_estimation')
+  );
--- a/supabase/migrations/20260313081705_migrate_cost_estimates_triggers_to_jwt.sql
+++ b/supabase/migrations/20260313081705_migrate_cost_estimates_triggers_to_jwt.sql
@@ -82,7 +82,7 @@ BEGIN
 
 
     IF NEW.is_locked THEN
-      NEW.locked_by_user_id := (SELECT id FROM users WHERE credential_id = (auth.jwt()->>'sub')::uuid);
+      NEW.locked_by_user_id := (auth.jwt()->'app_metadata'->>'internal_user_id')::uuid;
       NEW.locked_at := now();
     ELSE
       NEW.locked_by_user_id := NULL;
--- a/supabase/migrations/20260313081705_migrate_cost_estimates_triggers_to_jwt.sql
+++ b/supabase/migrations/20260313081705_migrate_cost_estimates_triggers_to_jwt.sql
@@ -1,0 +1,97 @@
+-- Update trigger function to use JWT claims instead of database lookups
+-- This improves performance by reading permissions from JWT tokens
+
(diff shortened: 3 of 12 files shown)


supabase/migrations/20260313081705_migrate_cost_estimates_triggers_to_jwt.sql
```diff
@@ -0,0 +1,97 @@
...
+  END IF;
+
+  -- Use JWT claims instead of database lookup
+  IF lock_changed THEN
+    IF NOT jwt_has_project_permission(
+      NEW.project_id,
+      'lock_cost_estimation'
+    ) THEN
+      RAISE EXCEPTION 'Insufficient permissions: lock_cost_estimation required to modify lock columns'
+        USING ERRCODE = '42501';
```
ripplearcgit, line 80: 🔴 **Blocking — DB lookup survives the JWT migration**

This line queries the `users` table on every lock operation:
```sql
NEW.locked_by_user_id := (SELECT id FROM users WHERE credential_id = (auth.jwt()->>'sub')::uuid);
```

`locked_by_user_id` holds `users.id` (internal app UUID), which differs from `auth.uid()` / `auth.jwt()->>'sub'` (the Supabase credential UUID) — so `auth.uid()` is not a drop-in replacement.

The correct fix: embed `users.id` in the JWT inside `custom_access_token_hook` alongside the existing `app_metadata.projects` claims, then read it here with no DB hit:

```sql
IF NEW.is_locked THEN
  NEW.locked_by_user_id := (auth.jwt()->'app_metadata'->>'internal_user_id')::uuid;
  NEW.locked_at := now();
```

This moves the DB lookup from every trigger fire to once at login — which is the goal of this entire PR.

ripplearcgit (changes requested):
## Review — `feat: migrate cost_estimates RLS policies and triggers to use JWT`

The RLS policy migration and test JWT claim fixes look correct. Two blocking issues before merge.

---

### 🔴 Blockers

**1. DB lookup survives in the trigger — contradicts the JWT migration goal**

The trigger still resolves `locked_by_user_id` via a database query:

```sql
NEW.locked_by_user_id := (SELECT id FROM users WHERE credential_id = (auth.jwt()->>'sub')::uuid);
```

This is a per-row DB hit on every lock operation — exactly what this PR is supposed to eliminate.

The reason `auth.uid()` alone isn't a drop-in fix is that `locked_by_user_id` stores `users.id` (the internal application UUID), while `auth.uid()` returns `users.credential_id` (the Supabase auth UUID). They are different values.

The correct JWT-based fix is to embed `users.id` in the JWT at token issuance time inside the `custom_access_token_hook` — the same place `app_metadata.projects` permissions are already being written. This is not a secret: JWT payloads are signed but not encrypted; the client can read the claims but cannot forge them. UUIDs are safe to include.

Add to the hook output:
```json
{
  "app_metadata": {
    "internal_user_id": "<users.id uuid>",
    "projects": { ... }
  }
}
```

Then read it in the trigger with no DB lookup:
```sql
IF NEW.is_locked THEN
  NEW.locked_by_user_id := (auth.jwt()->'app_metadata'->>'internal_user_id')::uuid;
  NEW.locked_at := now();
```

The DB lookup moves from every trigger fire to once at login — which is the whole point of the JWT migration.

---

**2. Migration files appear hand-authored**

Both migration files (`..._rls_to_jwt.sql`, `..._triggers_to_jwt.sql`) appear to be manually written rather than generated via `supabase db diff`. Project standard requires migrations to be generated from schema diffs to ensure the migration file and the schema source stay in sync. Please regenerate via `supabase db diff` and confirm schema source files are updated to match.

supabase/migrations/20260313081637_migrate_cost_estimates_rls_to_jwt.sql
```diff
@@ -0,0 +1,46 @@
...
+  );
+
+-- Policy for UPDATE operations (editing cost estimates)
+-- Users can update cost estimates if they have the 'edit_cost_estimation' permission in their JWT
+CREATE POLICY "cost_estimates_update_policy" ON "cost_estimates"
+  FOR UPDATE
+  TO authenticated
+  USING (
+    jwt_has_project_permission(project_id, 'edit_cost_estimation')
+  );
```
ripplearcgit, line 40: 🟡 **Non-blocking — `WITH CHECK` missing on UPDATE policy**

Postgres copies `USING` to `WITH CHECK` when omitted, so this is functionally correct today. Explicit `WITH CHECK` is clearer and guards against future relaxation of the immutable trigger.

```sql
CREATE POLICY "cost_estimates_update_policy" ON "cost_estimates"
  FOR UPDATE
  TO authenticated
  USING (
    jwt_has_project_permission(project_id, 'edit_cost_estimation')
  )
  WITH CHECK (
    jwt_has_project_permission(project_id, 'edit_cost_estimation')
  );
```

ripplearcgit (approved):
### 🟡 Non-blocking

**UPDATE policy missing `WITH CHECK`**

The UPDATE policy only has a `USING` clause. Postgres automatically copies `USING` to `WITH CHECK` when omitted, so this works correctly today — and the trigger's immutable guard on `project_id` means the attack vector (`WITH CHECK` catching a row mutated into a different project) is already blocked independently.

That said, writing `WITH CHECK` explicitly is better practice: it documents intent and is defensive against future changes to the immutable guard.

```sql
CREATE POLICY "cost_estimates_update_policy" ON "cost_estimates"
  FOR UPDATE
  TO authenticated
  USING (
    jwt_has_project_permission(project_id, 'edit_cost_estimation')
  )
  WITH CHECK (
    jwt_has_project_permission(project_id, 'edit_cost_estimation')
  );
```

---

### 🟢 Nit

**Test coverage gap: soft-delete permission boundary**

There is no test for a user who has `edit_cost_estimation` but not `delete_cost_estimation` attempting a soft-delete (`SET deleted_at = now()`). The trigger's delete branch would block this, but it isn't exercised. Minor gap worth closing in a follow-up.

SimonDerejeW:
> 2. Migration files appear hand-authored

Since npx supabase diff depends on a declarative schema setup, I’m sticking with manual migration generation via npx supabase migration until the underlying infrastructure (tables, RLS, and functions) is fully migrated.

supabase/migrations/20260313081301_auth_hook_project_permission_claims.sql
```diff
@@ -8,9 +8,15 @@ AS $$
 DECLARE
```
ripplearcgit, line 8: 🔴 **Blocking — do not edit existing applied migrations in-place**

This file has already been applied to provisioned environments. `supabase db push` skips files by timestamp, so this change will never run on any database that was set up before today.

The `internal_user_id` addition to the hook must be a new migration file with a later timestamp, e.g.:
```
20260412XXXXXX_add_internal_user_id_to_auth_hook.sql
```
Containing a `CREATE OR REPLACE FUNCTION custom_access_token_hook(...)` with the full updated body. Leave `20260313081301` untouched.

ripplearcgit (changes requested):
## Re-review — `feat: migrate cost_estimates RLS policies and triggers to use JWT`

Good progress. Three of four prior findings are resolved. Two blockers remain.

### ✅ Resolved
- **DB lookup in trigger** — `locked_by_user_id` now reads from `auth.jwt()->'app_metadata'->>'internal_user_id'` with no DB hit. Hook migration correctly embeds the claim at login. Hook test verifies the value.
- **UPDATE policy missing `WITH CHECK`** — explicit `WITH CHECK` now present on the UPDATE policy.
- **Missing soft-delete permission boundary test** — Test 11 added and covers the gap correctly.

---

### 🔴 Blockers

**1. Migration files appear hand-authored**

`20260313081637_migrate_cost_estimates_rls_to_jwt.sql` and `20260313081705_migrate_cost_estimates_triggers_to_jwt.sql` appear to be manually written. Project standard requires migration files to be generated via `supabase db diff` to ensure the migration and schema source stay in sync. Please regenerate both via `supabase db diff` and confirm the schema source files are updated to match.

**2. Hook change modifies an existing applied migration in-place**

`20260313081301_auth_hook_project_permission_claims.sql` has been edited directly to add the `internal_user_id` logic. This is incorrect: `supabase db push` is timestamp-driven and skips files it has already applied. Any environment that was provisioned before this push will never receive the `internal_user_id` change.

The hook update must be delivered as a new migration file with a later timestamp (e.g. `20260412XXXXXX_add_internal_user_id_to_auth_hook.sql`) containing a `CREATE OR REPLACE FUNCTION` for the hook. The existing `20260313081301` file should be left as-is.

supabase/migrations/20260413004017_add_internal_user_id_to_auth_hook.sql
```diff
@@ -0,0 +1,60 @@
...
+DECLARE
+  claims jsonb;
+  projects_claims jsonb;
+  user_internal_id uuid;
+BEGIN
+  claims := event->'claims';
+
+  -- Get the internal user ID (users.id) from credential_id
+  SELECT u.id INTO user_internal_id
+  FROM public.users u
```
ripplearcgit, line 19: 🟡 **Non-blocking — `user_internal_id` can be `NULL` silently**

If no `users` row matches the `credential_id`, `user_internal_id` remains `NULL` and gets embedded into the JWT as `"internal_user_id": null`. The trigger then casts this to `uuid` and writes `NULL` to `locked_by_user_id` on a lock operation.

Consider a guard:
```sql
IF user_internal_id IS NULL THEN
  RAISE WARNING 'custom_access_token_hook: no users row for credential_id %',
    (event->>'user_id')::uuid;
END IF;
```
This at least surfaces the misconfiguration in logs rather than letting it propagate silently into JWT claims.

ripplearcgit (approved):
## Re-review — `feat: migrate cost_estimates RLS policies and triggers to use JWT`

**Conditional LGTM** — all prior blockers resolved. One non-blocking finding below.

### ✅ Resolved
- **DB lookup in trigger** — `locked_by_user_id` reads from `auth.jwt()->'app_metadata'->>'internal_user_id'` with no DB hit.
- **In-place edit of applied migration** — hook change correctly delivered as new migration `20260413004017_add_internal_user_id_to_auth_hook.sql`.
- **UPDATE policy missing `WITH CHECK`** — explicit `WITH CHECK` present.
- **Missing soft-delete permission boundary test** — Test 11 added and covers the gap.

---

### 🟡 Non-blocking

**`user_internal_id` can be `NULL` if no matching `users` row exists**

In `20260413004017_add_internal_user_id_to_auth_hook.sql`, the `SELECT ... INTO user_internal_id` will silently leave `user_internal_id = NULL` if `credential_id` has no matching row. This value is then embedded into the JWT as `"internal_user_id": null`. When the trigger executes `(auth.jwt()->'app_metadata'->>'internal_user_id')::uuid`, it produces `NULL`, silently writing `NULL` to `locked_by_user_id` on a lock operation — undermining the field's intent.

Consider adding a guard after the SELECT:

```sql
SELECT u.id INTO user_internal_id
FROM public.users u
WHERE u.credential_id = (event->>'user_id')::uuid;

IF user_internal_id IS NULL THEN
  RAISE WARNING 'custom_access_token_hook: no users row for credential_id %',
    (event->>'user_id')::uuid;
  -- internal_user_id will be omitted or null in claims; lock operations will fail safely via trigger
END IF;
```

The trigger's lock branch will then produce a `NULL` `locked_by_user_id`, which depending on the column's `NOT NULL` constraint may raise an error — but that's silent from the hook's perspective. A warning log here at least surfaces the misconfiguration.

PR #34, opened 2026-05-07
Feat/init powersync

SimonDerejeW:
# PR Review: `feat/init-powersync` → `docs/power-sync-wiki-doc`

**Date:** Fri May 8, 2026
**Reviewer:** Automated Code Review
**PR Size Classification:** S (≈230 lines of production-relevant config/SQL — single, clear purpose ✅)

---

## 1. PR Summary

This PR introduces a self-hosted **PowerSync** integration alongside a Supabase backend. It wires up Docker Compose services (PowerSync + MongoDB replica set), environment configuration, sync rules, a PostgreSQL logical replication publication migration, and switches Supabase Auth from legacy symmetric JWT signing to asymmetric ES256 keys via a committed signing key file. The `README.md` added in `powersync/` is comprehensive and well-structured for onboarding developers. Overall, this is a focused and well-scoped initialisation PR — the declared `TODO: CA-647` items cleanly defer the broader entity sync work.

---

---

## 2. Review Summary Table

| Issue Name | Description | Status | Notes |
|---|---|---|---|
| **`latest` image tag for PowerSync** | `journeyapps/powersync-service:latest` in `compose.yaml` is non-deterministic and can break environments on image updates. Pin to a specific version tag. | ✅ | |
| **`log_level: debug` as default** | `service.yaml` defaults to `debug` logging, which is too verbose for shared dev environments and may expose sensitive replication data | ✅ | |
| **Redundant env var declarations** | `PS_POSTGRES_URI`, `PS_JWKS_URI`, and `PS_API_TOKEN` are declared twice in `compose.yaml` (once via `env_file`, once under `environment:`). Remove the duplicates under `environment:`. | ✅ |  |
| **README migration guidance** | README instructs developers to edit the existing migration to add tables. Supabase migrations are append-only; guidance should direct to creating a new migration using `ALTER PUBLICATION ... ADD TABLE`. | ✅ | |
| **`sslmode: disable` missing production warning** | The `service.yaml` replication connection disables SSL with no inline warning. A comment should clearly document that this is local-dev-only and must be changed for any non-local deployment. | ✅ |  |
| **No `wal_level` prerequisite comment in migration** | The SQL migration silently assumes `wal_level = logical`. A comment noting this prerequisite saves debugging time for developers on custom Postgres instances. |✅ | |

PR #35, opened 2026-05-08
CA-647: feat: configure PowerSync sync streams for user profiles, projects, a…

SimonDerejeW:
# PR Review: `feat/add-core-tables` → `feat/init-powersync` (Updated)

**Date:** Fri May 8, 2026
**Reviewer:** Automated Code Review
**PR Size Classification:** XS (≈58 lines — single, clear purpose ✅)

---

## 1. PR Summary

This is an updated revision of the core entity sync PR. The scope is identical to the previous iteration — expanding PowerSync sync streams to cover the authenticated user's profile (`my_user`), accessible projects (`user_projects`), and project memberships (`user_memberships`), alongside a proper append-only migration adding `users`, `projects`, and `project_members` to the logical replication publication.

**Change from previous review:** The `user_memberships` stream has been refactored from an inline correlated subquery to a `with:` CTE (`my_user_id`) — matching the pattern used by `user_projects`. This resolves the consistency and potential performance issue raised in the last review. ✅

---

---

## 3. Review Summary Table

|---|---|---|---|
| **Missing indexes for `user_projects` CTE** | Three-table join in `accessible_projects` CTE requires indexes on `users(credential_id)`, `project_members(user_id, membership_status)`, and `projects(project_status)`. Confirm these exist before merging. | ✅ |  |
| **`project_status != 'archived'` inequality filter** | Open-ended exclusion will silently sync any new project statuses added in the future. Replace with an explicit `IN ('active', 'draft')` allowlist. | ✅ |  |
| **No rollback comment in migration** | `ALTER PUBLICATION ... ADD TABLE` has no automatic rollback. Add a comment with the corresponding `DROP TABLE` command for operational safety. | ✅ |  |
| **`user_preferences` JSONB audit** | `user_preferences` is synced to client devices. Verify the JSONB field contains no server-only flags, admin metadata, or sensitive values that should not leave the server. | ✅ | |

### DIFF
--- a/powersync/README.md
+++ b/powersync/README.md
@@ -110,6 +110,59 @@ streams:
 
 Restart PowerSync after changes: `docker compose restart powersync`
 
+## Testing Sync Streams (without a frontend)
+
+You can verify sync rules and JWT-scoped data without building a client. Use the hosted
+[PowerSync Diagnostics App](https://diagnostics-app.powersync.com).
+
+### 1. Get a JWT from local Supabase
+
+Sign in against your local Supabase Auth to get an access token (replace credentials):
+
+```bash
+# Load Supabase anon key from your Supabase env
+SUPABASE_URL="http://localhost:54321"
+SUPABASE_ANON_KEY="<your local anon key from `npx supabase status`>"
+
+curl -X POST "$SUPABASE_URL/auth/v1/token?grant_type=password" \
+  -H "apikey: $SUPABASE_ANON_KEY" \
+  -H "Content-Type: application/json" \
+  -d '{"email":"you@example.com","password":"yourpassword"}' \
+  | jq -r .access_token
+```
+
+The returned JWT carries the `app_metadata.internal_user_id` claim injected by the
+`custom_access_token_hook` — that's what the sync rules filter on.
+
+### 2. Connect the Diagnostics App
+
+Open https://diagnostics-app.powersync.com and provide:
+
+- **Endpoint**: `http://localhost:${PS_PORT}` (read `PS_PORT` from your `.env`, default `8081`)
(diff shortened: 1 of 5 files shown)


ripplearcgit (approved):
## Conditional LGTM

The core sync stream design is correct — using `auth.parameter('$.app_metadata.internal_user_id')` to avoid a `users` table lookup on every sync evaluation is the right call given the JWT hook setup in #34. The CTE pattern in `user_projects` is consistent and clean. The append-only migration is properly structured with rollback guidance. CA-647 delivered.

Three non-blocking items to address before or shortly after merge:

---

### 🟡 N1 — Do not edit the committed migration `20260506000000`

`supabase/migrations/20260506000000_powersync_replication_setup.sql` is modified in this diff (the `-- TODO: CA-647` line is removed). Supabase migrations run exactly once — editing a file that has already been applied causes divergence between the migration history and the filesystem. The TODO being resolved by this PR is not a reason to touch the old file. Leave `20260506000000` untouched; the new work speaks for itself via `20260507235346` and the sync-config changes.

---

### 🟡 N2 — Confirm `user_preferences` is safe to sync

`my_user` stream explicitly selects `user_preferences` (JSONB). If this column stores any server-only flags, admin metadata, or values not intended for the client, they will land on every user's device. Please confirm in a follow-up comment what this field contains and that its full contents are client-safe.

---

### 🟡 N3 — `project_status = 'active'` implicit allowlist

The CTE filters `p.project_status = 'active'`. If `project_status` is a free-text column rather than a Postgres enum, future status values (e.g. `'paused'`, `'pending'`) will silently be excluded from sync with no warning. Either enforce `project_status` as a Postgres enum, or add a comment noting the intended set of allowed values.

PR #36, opened 2026-05-19
CA-687: Feat: Implement project search history module with RLS and RPC

Bazabizi:
## 1. PR SUMMARY

Adds the backend for the **Project Search** feature — a dedicated, dashboard-level search surface whose only target is projects. Ships `public.project_search_history` (user_id, search_term, has_results, search_count, timestamps), 4 personal-ownership RLS policies, two `BEFORE UPDATE` triggers reusing the global `search_history` trigger functions, and `public.get_project_search_suggestions(user_id)` RPC (SECURITY INVOKER, anti-spoof guarded, two-step personal→empty).

---

## 2. REVIEW SUMMARY TABLE

| Wrong feature scope (project-scoped vs Project Search) | Original implementation assumed searches happened *inside* a project (`project_id NOT NULL FK`, teammate visibility, RPC took `project_id` param). The actual feature is the Project Search dashboard surface — no project context exists at search time. | ✅ Addressed | Dropped `project_id` column, FK, and CASCADE; unique key now `(user_id, search_term)`; removed teammate SELECT policy; removed membership-gated INSERT; RPC signature is now `get_project_search_suggestions(user_id)` with two-step (personal → empty); README rewritten to reflect Project Search as a standalone feature isolated from Global Search; migration regenerated. |
| `CA-XXX` placeholder | README orphan-cleanup TODO had no real YouTrack ID (B2 from prior global_search PR review) | ✅ Addressed | Linked to `CA-597` |
| "Two-step" comment was previously wrong | RPC docstring said 2 steps on a 3-step algorithm | ✅ Addressed | After the rewrite the algorithm IS two-step (personal → empty), so the docstring now matches. |
| Wrong `db diff` command in README | `db diff -f` shadow-DB approach failed on cross-module FKs in the original design | ✅ Addressed | README documents the working `--from=migrations --to=local` flow; still applies even though the new design no longer has cross-module FKs (kept for parity with sibling modules). |
| Teammate JWT-policy follow-up | Original PR raised whether to migrate the teammate policy to `jwt_has_project_permission` | N/A | Teammate policy no longer exists — Project Search is personal-only. |

### DIFF
--- a/supabase/migrations/20260520112528_33_project_search_history.sql
+++ b/supabase/migrations/20260520112528_33_project_search_history.sql
@@ -1,0 +1,54 @@
+SET check_function_bodies = false;
+CREATE FUNCTION public.get_project_search_suggestions(user_id uuid)
+ RETURNS text[]
+ LANGUAGE plpgsql
+ SET search_path TO 'public'
+AS $function$
+DECLARE
+  v_suggestions text[];
+BEGIN
+
+  -- Anti-spoof guard: reject calls where the passed user_id does not
+  -- match the authenticated session. Prevents parameter spoofing via
+  -- the PostgREST API even though RLS would silently return empty rows.
+  IF user_id IS DISTINCT FROM auth.uid() THEN
+    RAISE EXCEPTION 'Unauthorized: user_id must match the authenticated session'
+      USING ERRCODE = '42501';
+  END IF;
+
+  -- Step 1: Personal history — own searches that returned results
+  SELECT ARRAY(
+    SELECT psh.search_term
+    FROM project_search_history psh
+    WHERE psh.user_id = auth.uid()
+      AND psh.has_results = true
+    ORDER BY psh.search_count DESC
+    LIMIT 10
+  ) INTO v_suggestions;
+
+  -- Step 2: empty array if nothing found
+  RETURN COALESCE(v_suggestions, ARRAY[]::text[]);
+
+END;
+$function$;
+COMMENT ON FUNCTION public.get_project_search_suggestions(uuid) IS 'Returns up to 10 search-term suggestions for the Project Search feature.
+Accepts user_id uuid — must match auth.uid() or raises 42501.
+Step 1: caller own history (has_results=true), ordered by frequency.
+Step 2: empty array if no personal history.
+SECURITY INVOKER — RLS on project_search_history enforces visibility.
+Does NOT expose users.credential_id or any auth identifier.
+has_results contract: caller must upsert with has_results=true after a non-empty result set.
+Rows with has_results=false are stored but excluded from suggestions.';
+CREATE TABLE public.project_search_history (id uuid DEFAULT gen_random_uuid() NOT NULL, user_id uuid NOT NULL, search_term character varying(255) NOT NULL, has_results boolean DEFAULT false NOT NULL, search_count integer DEFAULT 1 NOT NULL, created_at timestamp with time zone DEFAULT now() NOT NULL, updated_at timestamp with time zone DEFAULT now() NOT NULL);
+ALTER TABLE public.project_search_history ENABLE ROW LEVEL SECURITY;
+ALTER TABLE public.project_search_history ADD CONSTRAINT project_search_history_pkey PRIMARY KEY (id);
+CREATE INDEX project_search_history_updated_at_idx ON public.project_search_history (updated_at);
+CREATE UNIQUE INDEX project_search_history_user_term_uq ON public.project_search_history (user_id, search_term);
+CREATE INDEX project_search_history_user_id_idx ON public.project_search_history (user_id);
+CREATE INDEX project_search_history_has_results_idx ON public.project_search_history (has_results) WHERE has_results = true;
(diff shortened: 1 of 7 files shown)


supabase/migrations/20260520112528_33_project_search_history.sql
```diff
@@ -0,0 +1,54 @@
...
+Does NOT expose users.credential_id or any auth identifier.
+has_results contract: caller must upsert with has_results=true after a non-empty result set.
+Rows with has_results=false are stored but excluded from suggestions.';
+CREATE TABLE public.project_search_history (id uuid DEFAULT gen_random_uuid() NOT NULL, user_id uuid NOT NULL, search_term character varying(255) NOT NULL, has_results boolean DEFAULT false NOT NULL, search_count integer DEFAULT 1 NOT NULL, created_at timestamp with time zone DEFAULT now() NOT NULL, updated_at timestamp with time zone DEFAULT now() NOT NULL);
+ALTER TABLE public.project_search_history ENABLE ROW LEVEL SECURITY;
+ALTER TABLE public.project_search_history ADD CONSTRAINT project_search_history_pkey PRIMARY KEY (id);
+CREATE INDEX project_search_history_updated_at_idx ON public.project_search_history (updated_at);
+CREATE UNIQUE INDEX project_search_history_user_term_uq ON public.project_search_history (user_id, search_term);
+CREATE INDEX project_search_history_user_id_idx ON public.project_search_history (user_id);
+CREATE INDEX project_search_history_has_results_idx ON public.project_search_history (has_results) WHERE has_results = true;
```
ripplearcgit, line 45: 🔴 **Blocking** — `CREATE TRIGGER` here will error with "trigger already exists" on any re-run (`supabase db reset`, CI double-apply, etc.). `04_triggers.sql` correctly uses `CREATE OR REPLACE TRIGGER` — the migration must match.

Fix: regenerate via `supabase db diff --from=migrations --to=local`, or replace both `CREATE TRIGGER` lines with `CREATE OR REPLACE TRIGGER`.

ripplearcgit (changes requested):
## Round 2 Review

Overall direction is correct — personal-only table, clean isolation from Global Search, reused trigger functions, correct SECURITY INVOKER + anti-spoof guard, no `credential_id` leakage. One blocker before merge.

---

### 🔴 Blocking — Migration uses `CREATE TRIGGER`, schema uses `CREATE OR REPLACE TRIGGER`

**File:** `supabase/migrations/20260520112528_33_project_search_history.sql` (lines 48–51)

The schema source `04_triggers.sql` correctly uses `CREATE OR REPLACE TRIGGER` for both triggers. The generated migration uses plain `CREATE TRIGGER` (no `OR REPLACE`).

On any re-run — e.g. `supabase db reset` after a failed first attempt, or CI applying migrations twice — this will throw **"trigger already exists"** and halt the migration chain.

**Fix:** Regenerate the migration via `supabase db diff --from=migrations --to=local` as documented in the README, or manually replace both `CREATE TRIGGER` statements with `CREATE OR REPLACE TRIGGER` in the migration file.

```sql
-- current (migration) — will error on re-run
CREATE TRIGGER trigger_increment_project_search_count ...
CREATE TRIGGER trigger_set_project_search_history_updated_at ...

-- required — matches 04_triggers.sql
CREATE OR REPLACE TRIGGER trigger_increment_project_search_count ...
CREATE OR REPLACE TRIGGER trigger_set_project_search_history_updated_at ...
```

supabase/schemas/global_search/project_search_history/03_functions.sql
```diff
@@ -0,0 +1,87 @@
...
+    RAISE EXCEPTION 'Unauthorized: user_id must match the authenticated session'
+      USING ERRCODE = '42501';
+  END IF;
+
+  -- Step 1: Personal history — own searches that returned results
+  SELECT ARRAY(
+    SELECT psh.search_term
+    FROM project_search_history psh
+    WHERE psh.user_id = auth.uid()
+      AND psh.has_results = true
```
ripplearcgit, line 68: 🟡 **Nit** — `COALESCE(v_suggestions, ARRAY[]::text[])` is dead code. `SELECT ARRAY(subquery)` never returns `NULL` in PostgreSQL — it returns `'{}'::text[]` when the subquery is empty. The fallback branch can never fire. Either return `v_suggestions` directly or inline the subquery and drop the variable.

ripplearcgit (approved):
### 🟡 Nit — `COALESCE` on `SELECT ARRAY(…)` is a no-op

**File:** `supabase/schemas/global_search/project_search_history/03_functions.sql` (line 68)

In PostgreSQL, `SELECT ARRAY(subquery)` **never** returns `NULL` — when the subquery has no rows it returns `'{}'::text[]` (an empty array). So `COALESCE(v_suggestions, ARRAY[]::text[])` can never take the fallback branch. The Step 2 comment ("empty array if nothing found") implies a null-check is happening, which is misleading.

```sql
-- Option A: inline the subquery, drop the variable entirely
RETURN ARRAY(
  SELECT psh.search_term
  FROM project_search_history psh
  WHERE psh.user_id = auth.uid()
    AND psh.has_results = true
  ORDER BY psh.search_count DESC
  LIMIT 10
);

-- Option B: keep the variable, just drop the COALESCE
SELECT ARRAY(...) INTO v_suggestions;
RETURN v_suggestions;  -- already '' if no rows
```

SimonDerejeW (approved):

PR #38, opened 2026-05-21
CA-648 feat: add user_cost_estimates sync stream and corresponding mi…

SimonDerejeW:
# PR Review: feat/cost-estimaation-syncc → master

**Reviewed by:** Claude  
**Date:** Fri May 22 2026  
**PR Size:** ~60 lines of production code → **S (Small)** ✅

---

---

## 2. REVIEW SUMMARY TABLE

|---|---|---|---|
| Non-idempotent publication migration | `ALTER PUBLICATION powersync ADD TABLE` will error if any table is already in the publication. No `IF NOT EXISTS` exists in Postgres for this statement. |  | Idempotency is guaranteed by Supabase's migration runner `supabase_migrations.schema_migrations` this SQL runs exactly once per environment. |
| Column list not verified against schema | The explicit column SELECT in `sync-config.yaml` should be cross-checked against `01_table.sql` to confirm no columns have been added since the list was written.  | ✅ |  |

### DIFF
--- a/powersync/sync-config.yaml
+++ b/powersync/sync-config.yaml
@@ -56,3 +56,38 @@ streams:
              joined_at, membership_status
       FROM project_members
       WHERE user_id = auth.parameter('$.app_metadata.internal_user_id')
+
+  # ── 4. COST ESTIMATES ───────────────────────────────────────────
+  # Gated by the 'get_cost_estimations' permission to mirror the RLS
+  # policy in 20260313081637_migrate_cost_estimates_rls_to_jwt.sql.
+  # PowerSync bypasses Postgres RLS, so the sync rule MUST enforce
+  # the same permission check the API enforces — otherwise users
+  # without the permission would still receive the rows.
+  #
+  # Soft-deleted rows (deleted_at IS NOT NULL) are excluded.
+  user_cost_estimates:
+    auto_subscribe: false  # on-demand: app subscribes when user enters cost estimation feature to avoid evaluating the 4-table CTE join on every connection startup
+    with:
+      cost_estimate_projects: |
+        SELECT DISTINCT pm.project_id AS id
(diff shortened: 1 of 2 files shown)


ripplearcgit (approved):
## Conditional LGTM ✅

Clean, well-documented infrastructure PR. The permission-gate design is correct — the `cost_estimate_projects` CTE faithfully mirrors the JWT-based RLS policy from `20260313081637_migrate_cost_estimates_rls_to_jwt.sql`, and publishing `role_permissions` / `permissions` without exposing them in a sync stream is the right pattern for CTE-only lookup tables.

Column list cross-checked against `20250514093710_16_cost_estimates.sql` and `20260126000000_soft_delete_cost_estimates.sql` — all 20 non-audit columns present; `deleted_at` correctly excluded from SELECT and used only in the WHERE filter.

Migration verified clean — all three tables (`cost_estimates`, `role_permissions`, `permissions`) confirmed absent from the publication prior to this migration.

---

### 🟡 Non-blocking — `powersync/sync-config.yaml` line 66

**`auto_subscribe: true` — monitor replicator load post-rollout.**

With `auto_subscribe: true`, PowerSync evaluates the `cost_estimate_projects` CTE (a 4-table join across `project_members → projects → role_permissions → permissions`) for every authenticated connection on startup. For small teams this is fine; at scale it could generate sustained load on the replicator. Consider whether this stream should remain auto-subscribed long-term or be triggered on-demand by the app when the user enters the cost estimation feature. No action required before merge — just worth monitoring after rollout in staging and production.

---

The inline comment block explaining why PowerSync bypasses Postgres RLS and why the sync rule must therefore replicate the same permission check is excellent — exactly the kind of documentation that prevents future data-leak regressions. 👍

PR #39, opened 2026-06-11
CA-252: fix(rls): add public SELECT policy for tags reference table

Bazabizi:
## Summary

The `tags` table (created in `20250514010735_06_tags.sql`) has RLS enabled via `20250514131510_enable_rls.sql` but **no policy was ever defined**, so authenticated client reads silently return zero rows.

The Flutter app is about to start reading this table — ripplearc/construculator-app#345 (CA-252, `TagRepository`) and ripplearc/construculator-app#346 (CA-638, Tags filter sheet) — and without this policy the tag filter UI will always show an empty list.

This adds a public SELECT policy following the exact `professional_roles_select_public` reference-data pattern: read-only for clients, writes restricted to service_role/migrations.

## Changes

- `supabase/schemas/core/tags/01_rls.sql` — declarative policy file
- `supabase/migrations/<ts>_34_tags_select_rls_policy.sql` — migration

## Testing

Policy mirrors `professional_roles` (`supabase/schemas/core/professional_roles/03_rls.sql`). After applying, `select * from tags` as an authenticated user returns rows.

## CI note

> [!NOTE]
> The `database_tests` failure on this PR is **not caused by these changes** — it's a breaking change in Supabase CLI v2.106.0 (released after the last green build) that revokes the default Data API grants our schema relied on. Confirmed via no-op control #40; root cause and fix in #41 (stopgap) / [CA-729](https://ripplearc.youtrack.cloud/issue/CA-729) (permanent explicit GRANTs). Re-run checks after #41 merges.

### DIFF
--- a/supabase/migrations/20260611223604_34_tags_select_rls_policy.sql
+++ b/supabase/migrations/20260611223604_34_tags_select_rls_policy.sql
@@ -5,6 +5,7 @@
 -- Public SELECT Policy
 -- Tags are reference data used for search filtering; anyone can read them.
 
+DROP POLICY IF EXISTS "tags_select_public" ON "public"."tags";
 CREATE POLICY "tags_select_public" ON "public"."tags" FOR SELECT USING (true);
 
 -- No INSERT/UPDATE/DELETE policies defined.
--- a/supabase/migrations/20260611223604_34_tags_select_rls_policy.sql
+++ b/supabase/migrations/20260611223604_34_tags_select_rls_policy.sql
@@ -1,0 +1,12 @@
+-- Tags RLS Policies
+
+-- RLS is already enabled for "tags" (see 20250514131510_enable_rls.sql).
+
+-- Public SELECT Policy
+-- Tags are reference data used for search filtering; anyone can read them.
+
+CREATE POLICY "tags_select_public" ON "public"."tags" FOR SELECT USING (true);
+
+-- No INSERT/UPDATE/DELETE policies defined.
+-- Write access is intentionally restricted to service_role / migrations only,
+-- matching the professional_roles reference-data pattern.
--- a/supabase/schemas/core/tags/01_rls.sql
+++ b/supabase/schemas/core/tags/01_rls.sql
@@ -1,0 +1,12 @@
+-- Tags RLS Policies
+
+-- RLS is already enabled for "tags" (see 20250514131510_enable_rls.sql).
+
+-- Public SELECT Policy
+-- Tags are reference data used for search filtering; anyone can read them.
+
+CREATE POLICY "tags_select_public" ON "public"."tags" FOR SELECT USING (true);
+
+-- No INSERT/UPDATE/DELETE policies defined.
+-- Write access is intentionally restricted to service_role / migrations only,
+-- matching the professional_roles reference-data pattern.
--- a/supabase/schemas/core/tags/01_table.sql
+++ b/supabase/schemas/core/tags/01_table.sql
@@ -1,0 +1,23 @@
+-- Tags Table
+-- Reference data used for search filtering
+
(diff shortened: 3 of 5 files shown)


ripplearcgit (approved):
## Summary

Fixes a silent default-deny gap on the `tags` reference table — RLS was enabled (via `20250514131510_enable_rls.sql`) but no SELECT policy was ever added, causing all authenticated reads to return an empty list. This PR adds `tags_select_public` using the established `professional_roles_select_public` pattern.

**Non-blocking nits:**
- 🟡 No `IF NOT EXISTS` / `DROP POLICY IF EXISTS` guard on `CREATE POLICY` — consistent with existing pattern but worth hardening in a future pass.
- 🟡 `supabase/schemas/core/tags/` only contains `01_rls.sql`; the table/index declarative files are missing from the directory (they exist only as older migrations). Consider backfilling to match the `professional_roles` module structure.

## Verdict

✅ **APPROVE (Conditional LGTM)**

RLS fix is correct, minimal, and follows the established reference-data pattern. CI failure is the pre-existing Supabase CLI v2.106.0 regression addressed in #41 — not caused by these changes. Merge after #41 lands and checks go green.

**Merge order:** `#41 (CI fix) → #39 (this) → app#345 (TagRepository) → app#346 (filter sheet)`

ripplearcgit (approved):
### Architectural overview

```mermaid
sequenceDiagram
    participant Flutter as Flutter App
    participant Supabase as Supabase Client
    participant Auth as Supabase Auth
    participant DB as PostgreSQL
    participant RLS as RLS Engine

    Flutter->>Supabase: supabase.from('tags').select()
    Supabase->>Auth: Resolve session (anon or authenticated)
    Auth->>DB: Forward query
    DB->>RLS: Evaluate SELECT policies on tags
    alt Before this PR (no SELECT policy)
        RLS->>DB: Default-deny: no rows returned
        DB->>Flutter: [] (empty list, silent failure)
    else After this PR (tags_select_public USING true)
        RLS->>DB: Allow all rows
        DB->>Flutter: [tag1, tag2, tag3, ...]
    end
```

### Merge chain

```mermaid
flowchart LR
    A["PR #41 — CI Fix\nRestore Data API grants"] -->|unblocks| B["PR #39\ntags SELECT RLS\nyou are here"]
    B -->|unblocks| C["App PR #345\nTagRepository"]
    C -->|unblocks| D["App PR #346\nTags filter sheet"]
    style A fill:#dafbe1
    style B fill:#0969da,color:#fff
    style C fill:#e8f0fe
    style D fill:#e8f0fe
```

The RLS fix is correct, minimal, and follows the established `professional_roles` reference-data pattern. Safe to merge once PR #41 lands and CI goes green.

🟡 **Nit: no `IF NOT EXISTS` guard on `CREATE POLICY`**
*(not in diff — suggestion for future policy migrations)*

The migration uses a bare `CREATE POLICY "tags_select_public" ...`. If run twice (e.g. a dev reset that partially applied this migration) it would error with *"policy already exists"*. Consistent with `professional_roles` but worth noting as a minor hardening opportunity. Add `DROP POLICY IF EXISTS "tags_select_public" ON "public"."tags";` before the `CREATE POLICY`, or use `CREATE POLICY IF NOT EXISTS` (Postgres 17+).

🟡 **Nit: `supabase/schemas/core/tags/` only has `01_rls.sql`**
*(structural gap — not in diff)*

The `professional_roles` schema directory has `01_table.sql`, `02_indexes.sql`, `03_rls.sql`, `04_triggers.sql`, and a `README.md`. Consider backfilling the remaining schema files in a follow-up to keep the declarative schema directory consistent with the module pattern.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-backend-39.html

PR #42, opened 2026-06-24
CA-752: fix(search): auto-update projects.updated_at on edits

Bazabizi:
## Summary
- `projects.updated_at` was never bumped on `UPDATE` — there was no trigger maintaining it, unlike `cost_estimates`. Editing a project's name/description left `updated_at` frozen at creation time.
- Deploys `public.set_current_timestamp_updated_at()`, a shared trigger function already documented in `supabase/schemas/_shared/01_functions.sql` for reuse across tables (`users`, `companies`, `professional_roles` already reference it in their schema files) but never actually migrated, and attaches it to `projects` via `CREATE OR REPLACE TRIGGER`.
- Adds the matching `supabase/schemas/core/projects/04_triggers.sql` schema file and a `core/projects/README.md` documenting the trigger (per this repo's schema-module-doc rule; table/indexes/RLS for `projects` are not yet backfilled into `schemas/` — out of scope here, noted in the README).
- This is a prerequisite for CA-170's global search "Modified" date filter to mean anything for projects (follow-up PR #43), but is a standalone, independently-reviewable bug fix.

## Test plan
- [x] `supabase test db` — new `projects_updated_at_trigger_test.sql` passes (asserts `updated_at` is bumped past `now() - 1 minute` after an `UPDATE`, and that it no longer equals `created_at`).
- [x] Full local pgTAP suite passes with this migration applied in isolation (145/145).

## Review Summary

| Issue Name | Description | Status | Notes |
|---|---|---|---|
| Schema doc backfill | New trigger touched `projects`, which had no `supabase/schemas/core/projects/` module or README yet | ✅ Addressed | Added `04_triggers.sql` + `README.md` scoped to the trigger; full table/indexes/RLS backfill left as a separate follow-up (noted in the README) |

PR #43, opened 2026-06-24
CA-752: feat(search): support modified-date range filter on global_search RPC

Bazabizi:
## Summary
- Replaces `global_search`'s single `filter_by_date` param with `filter_by_date_from`/`filter_by_date_to` (both `timestamptz`, `DEFAULT NULL`) to match the Flutter date-range filter shipped in CA-170/CA-377 (`DateFilterChip`, presets, custom range picker).
- Applies both bounds as an inclusive range on `updated_at` for **both** `projects` and `cost_estimates` — the design doc mentions only projects, but the function's own `SELECT` targets show estimations are filtered too, so that behavior is preserved (just switched from `created_at` to `updated_at`, matching "Modified" semantics).
- Rejects an inverted range (`filter_by_date_from > filter_by_date_to`) with `RAISE EXCEPTION ... USING ERRCODE = '22023'` instead of silently returning empty results, so the Flutter client can surface a validation message.
- Function stays `LANGUAGE plpgsql` with no explicit `SECURITY DEFINER`/`INVOKER` change — unchanged from before, still relies on RLS for row visibility.
- Adds the matching `supabase/schemas/global_search/global_search/03_functions.sql` schema file and a `README.md` for the new module (per this repo's schema-module-doc rule).
- Depends on #42 (`projects.updated_at` trigger) so the new filter reflects real edits, not just creation time, for projects.

**Param naming:** `filter_by_date_from`/`filter_by_date_to` — confirmed this matches the Flutter `_toRpcParams` mapping already shipped (`remote_global_search_data_source.dart`), no client-side change needed for the CA-170 feature.

**Note:** Traced two *other* Flutter call sites that also invoke `global_search` (`lib/libraries/project/data/data_source/remote_project_search_data_source.dart` and `lib/features/dashboard/data/data_source/remote_project_search_data_source.dart`) — both send the old `filter_by_date`/`offset` params, which were already invalid against `global_search`'s signature on `master` before this PR (the real param names have always been `projects_offset`/`estimations_offset`/`members_offset`, never a single `offset`). Confirmed neither class is actually wired into the app (no DI registration reaches either one, no bloc/page calls them) — dead code, not a live regression. Not fixed here; flagging as a separate cleanup candidate.

## Test plan
- [x] `supabase test db` — `global_search_date_range_filter_test.sql` covers: no bounds (= today's no-filter behavior), `from` only, `to` only, both bounds, and the inverted-range error.
- [x] Full local pgTAP suite passes with both migrations applied in order (151/151).

## Review Summary

|---|---|---|---|
| PR size (Rule 1) | Raw diff is ~470 lines. Production code (migration `.sql` + schema mirror `.sql`, excluding README/tests per Rule 1's own "Production Code Only" scope) is 312 lines — still over the "M" threshold | ❌ Disagree (with reasoning, see breakdown below) | Not splittable: the 312 lines are `CREATE OR REPLACE FUNCTION` restating the **entire** function body in both the migration and its schema-file mirror (this repo's existing pairing convention, e.g. `project_search_history`), most of which is the unchanged `projects`/`estimations`/`members` query blocks carried forward verbatim — not new logic to review twice. The actual new/changed logic (2 new params, the inverted-range check, swapping `created_at`→`updated_at` in two `WHERE` clauses) is ~20 lines. Splitting further isn't possible without separating the migration from the schema backfill Rule 7 requires alongside it in the same PR. |
| Schema doc backfill | New `global_search/global_search/` schema directory had no README | ✅ Addressed | Added `README.md` documenting the RPC's full param table, the date-range semantics, and the source-of-truth diff workflow, matching the sibling `project_search_history`/`search_history` modules |


ripplearcgit (approved):
## CA-752: feat(search): support modified-date range filter on global_search RPC

**Verdict: Conditional LGTM** - 0 blockers, 2 nits. Pre-merge: merge #42 first, then retarget this PR base to master.

### What's right

- **Old 9-param overload correctly dropped** before installing the new 10-param one.
- **Validation-first guard order**: scope check then inverted-range check then queries; invalid inputs fail fast with clear 22023 errors.
- **SECURITY INVOKER is correct** - no privilege escalation needed; existing RLS handles row visibility.
- **credential_id privacy note preserved** in the schema doc.
- **pgTAP 151/151** - inverted-range rejection, no-bounds, from-only, to-only, and both-bounds cases all exercised.

---

### Nit 1 - pgTAP test fixtures cover projects only; cost_estimates date-filter path never exercised

supabase/tests/functions/global_search_date_range_filter_test.sql inserts two projects rows with backdated updated_at values and verifies all five filter scenarios. No cost_estimates fixtures are inserted, so the estimations array is always empty in every assertion - the date-filter logic on cost_estimates is never actually tested.

The SQL path is identical to the projects case so the risk is low, but a follow-up fixture addition would close the gap.

---

### Nit 2 - cost_estimates.updated_at maintained via permission-guard trigger, not a dedicated function

PR #42 gave projects a dedicated set_current_timestamp_updated_at() trigger. cost_estimates.updated_at is set as a side-effect inside check_cost_estimate_update_permissions. This works correctly for authenticated-role UPDATEs today, but a future service-role UPDATE would leave cost_estimates.updated_at stale while projects.updated_at auto-updates.

Recommend noting this asymmetry in the core/projects/README.md so the next engineer doing a service-role migration knows to set it manually.

---
Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-backend-43.html

PR #44, opened 2026-07-06
CA-737: feat(search): support multi-owner filter on global_search RPC

Bazabizi:
## Summary

Backend half of [CA-737](https://ripplearc.youtrack.cloud/issue/CA-737): the Flutter owner filter sheet allows multi-select, but `global_search` only accepted a single `filter_by_owner uuid`, so `GlobalSearchBloc` forwards just the alphabetically-first selected owner as a stopgap.

- **Migration 37** (generated via `supabase db diff --from=migrations --to=local` from the declarative schema file, per the module README workflow): drops the old `filter_by_owner uuid` overload (exact signature) and recreates `public.global_search` with `filter_by_owners uuid[] DEFAULT NULL`, re-granting EXECUTE to `anon`/`authenticated`/`service_role` explicitly (DROP removes prior grants). The predicate applies to both `projects` and `cost_estimates`: `creator_user_id = ANY(filter_by_owners)`.
- **Contract**: `NULL` **and an empty array** both mean "no owner filter" — an empty array is the natural "no owners selected" client state and must not silently match zero rows (`= ANY('{}')` is always false without the guard).
- **Schema backfill**: `global_search/global_search/03_functions.sql` mirrors the migration body exactly; module README parameter table updated.
- **Drift cleanup**: removed a stale pre-CA-752 duplicate definition of `global_search` (old `filter_by_date`/`filter_by_owner` signature) that was still sitting in `search_history/03_functions.sql` + its README section, so the declarative schema dir has exactly one definition per function. The CA-598 performance TODO it carried was moved to the canonical module README.

### Breaking change

The single-`filter_by_owner` signature is dropped (same intentional pattern as CA-752's `filter_by_date` removal). App-side callers must migrate to `filter_by_owners` — tracked by the remaining Flutter acceptance criteria on CA-737.

## Test plan

- New `global_search_multi_owner_filter_test.sql` (11 pgTAP assertions): NULL / empty-array / single-owner / two-owner filtering on **both** the projects and cost_estimates paths, owner×date-range intersection on both paths, and a `has_function` pin of the new signature. Runs under `SET LOCAL ROLE authenticated` with full JWT `app_metadata.projects` claims for cost_estimates RLS.
- Relaxed `global_search_date_range_filter_test.sql`'s `has_function` to a name-only existence check: the exact 10-parameter signature is pinned only by the multi-owner suite (the newest signature-changing test), so future signature changes won't break unrelated suites.
- Verified locally with the CI-pinned Supabase CLI 2.106.0: `supabase db reset` applies migration 37 cleanly on a fresh DB; `supabase test db` passes 165/165 across all 15 suites; psql smoke call with `filter_by_owners => ARRAY[...]` returns the expected shape with no `credential_id` exposure.

## Review Summary

All review findings are resolved — no outstanding issues in this PR.

| Stale duplicate `global_search` definition | `search_history/03_functions.sql` still carried the pre-CA-752 9-arg definition (`filter_by_date`, `filter_by_owner`); applying schema files to a fresh DB would create a second ambiguous overload. Its README also documented the removed params. | ✅ Addressed | Removed in this PR; pointer comments left; CA-598 perf TODO moved to the canonical `global_search` module README. |
| Old-signature callers | Dropping `filter_by_owner uuid` means clients still sending it get a function-not-found error. | ✅ Addressed | Intentional, documented breaking change (see Breaking change section) — same pattern as CA-752's `filter_by_date` removal. No backend action remains; the app-side rename ships with CA-737's Flutter half. |


ripplearcgit (approved):
✅ **All checks pass — ready to merge**

Migration 37 safely replaces single-owner filter with multi-owner array. NULL/empty-array contract explicitly guarded on both projects and cost_estimates paths. RLS remains deferred to table policies. 11 pgTAP assertions verify all filtering combinations and RLS enforcement. Function signature pinned; stale duplicate removed.

This is the parent PR: app-side PR #407 (CA-737 Flutter half) must not ship before this migration deploys.

**Full report:** ~/Desktop/pr-reviews/pr-review-construculator-backend-44.html

PR #45, opened 2026-07-14
CA-597: feat(search): purge orphaned search_history rows via pg_cron

Bazabizi:
## What

Adds a periodic cleanup that deletes orphaned rows from `search_history` and `project_search_history`. Both tables store `auth.uid()` in `user_id` with **no FK** to `auth.users` (cross-schema boundary), so rows are never cascade-deleted when a user is removed from Supabase Auth and accumulate as orphans (CA-597).

## How

- **`public.purge_orphaned_search_history()`** — `SECURITY DEFINER`, `search_path` pinned to `public, auth`. Anti-joins (`NOT EXISTS`) both tables against `auth.users(id)` and deletes rows with no matching auth user. `SECURITY DEFINER` is needed to read `auth.users` and to bypass the per-user RLS `DELETE` policies.
- **Lockdown** — `EXECUTE` revoked from `PUBLIC`/`anon`/`authenticated`, granted only to `postgres`. As defense-in-depth, the body rejects any call carrying `request.jwt.claims` (which PostgREST always sets and `pg_cron` never does) with `42501`, so it stays uncallable via the Data API even if `auto_expose_new_tables` re-grants `EXECUTE` on `db reset` (see `config.toml` / CA-729).
- **Schedule** — `pg_cron` job `purge-orphaned-search-history`, daily at 03:00 UTC, registered idempotently (unschedule-if-exists → schedule).
- **Scope** — covers both `search_history` and `project_search_history`; the latter's README already pointed CA-597 at it.
- **Docs** — the Orphan-Risk `TODO (CA-597)` in both module READMEs is replaced with an implemented-solution description; new `search_history_cleanup` module README added.

## Layout

New declarative module `supabase/schemas/global_search/search_history_cleanup/` (`03_functions.sql`, `07_cron.sql`, `README.md`) + generated migration `20260714230219_38_search_history_cleanup.sql`. The migration is hand-authored and annotated because `pg_cron` objects live in the `cron` schema, which `supabase db diff` does not track; the `07_cron.sql` statements remain the source of truth.

## Verification

pgTAP test `supabase/tests/functions/purge_orphaned_search_history_test.sql` (6 tests): asserts orphan rows are purged from **both** tables, active-user rows are untouched in both, and the API guard rejects a call carrying `request.jwt.claims`. All 6 pass locally; the migration applies cleanly and idempotently.

### DIFF
--- a/supabase/migrations/20260714230219_38_search_history_cleanup.sql
+++ b/supabase/migrations/20260714230219_38_search_history_cleanup.sql
@@ -27,7 +27,7 @@ SET search_path = public, auth
 AS $$
 BEGIN
   -- Defense-in-depth against an authenticated Data API call. Primary access
-  -- control is the REVOKE below (anon/authenticated hold no EXECUTE); this
+  -- control is the REVOKE below (no Data API role holds EXECUTE); this
   -- guard is a second layer in case the CLI's auto_expose_new_tables grant
   -- pass re-grants EXECUTE on db reset (a repo-wide issue tracked by CA-729,
   -- affecting every function). PostgREST populates request.jwt.claims with a
@@ -64,6 +64,7 @@ purge-orphaned-search-history pg_cron job (CA-597). Not exposed via the API.';
 REVOKE EXECUTE ON FUNCTION public.purge_orphaned_search_history() FROM PUBLIC;
 REVOKE EXECUTE ON FUNCTION public.purge_orphaned_search_history() FROM anon;
 REVOKE EXECUTE ON FUNCTION public.purge_orphaned_search_history() FROM authenticated;
+REVOKE EXECUTE ON FUNCTION public.purge_orphaned_search_history() FROM service_role;
 GRANT EXECUTE ON FUNCTION public.purge_orphaned_search_history() TO postgres;
 
 -- ============================================================
--- a/supabase/migrations/20260714230219_38_search_history_cleanup.sql
+++ b/supabase/migrations/20260714230219_38_search_history_cleanup.sql
@@ -1,0 +1,83 @@
+-- CA-597: Purge orphaned search-history rows.
+-- search_history.user_id and project_search_history.user_id store auth.uid()
+-- with no FK to auth.users (cross-schema boundary), so rows are not
+-- cascade-deleted when a user is removed from Supabase Auth. This adds a
+-- SECURITY DEFINER purge function and a daily pg_cron job that deletes rows
+-- whose user_id no longer exists in auth.users, from both tables.
+-- https://ripplearc.youtrack.cloud/issue/CA-597
+--
+-- Hand-authored (not `supabase db diff`-generated): pg_cron objects live in
+-- the `cron` schema, which `db diff` does not track. The executable
+-- statements below mirror supabase/schemas/global_search/
+-- search_history_cleanup/ (03_functions.sql + 07_cron.sql), which remain
+-- the source of truth; only migration boilerplate (check_function_bodies)
+-- and some explanatory comments differ.
+-- (Annotated per the CA-737 precedent for annotated migrations.)
+
+set check_function_bodies = off;
+
+-- ============================================================
+-- Purge function (see search_history_cleanup/03_functions.sql).
+-- ============================================================
+CREATE OR REPLACE FUNCTION public.purge_orphaned_search_history()
+RETURNS void
+LANGUAGE plpgsql
+SECURITY DEFINER
+SET search_path = public, auth
+AS $$
+BEGIN
+  -- Defense-in-depth against an authenticated Data API call. Primary access
+  -- control is the REVOKE below (anon/authenticated hold no EXECUTE); this
+  -- guard is a second layer in case the CLI's auto_expose_new_tables grant
+  -- pass re-grants EXECUTE on db reset (a repo-wide issue tracked by CA-729,
+  -- affecting every function). PostgREST populates request.jwt.claims with a
+  -- non-empty JWT for an authenticated request; pg_cron and direct
+  -- service/psql calls never set it. This is a maintenance function that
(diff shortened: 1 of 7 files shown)


supabase/migrations/20260714230219_38_search_history_cleanup.sql
```diff
@@ -0,0 +1,83 @@
...
+$$;
+
+COMMENT ON FUNCTION public.purge_orphaned_search_history IS
+'Deletes search_history and project_search_history rows whose user_id no
+longer exists in auth.users. SECURITY DEFINER; invoked by the daily
+purge-orphaned-search-history pg_cron job (CA-597). Not exposed via the API.';
+
+REVOKE EXECUTE ON FUNCTION public.purge_orphaned_search_history() FROM PUBLIC;
+REVOKE EXECUTE ON FUNCTION public.purge_orphaned_search_history() FROM anon;
+REVOKE EXECUTE ON FUNCTION public.purge_orphaned_search_history() FROM authenticated;
```
ripplearcgit, line 66: 🟡 **REVOKE list omits `service_role`**

The migration REVOKEs EXECUTE from `PUBLIC`, `anon`, and `authenticated`, and GRANTs to `postgres` — but never mentions `service_role`. Because `config.toml` keeps `auto_expose_new_tables = true` (the CA-729 workaround), a `db reset` re-grants EXECUTE to all three Data-API roles including `service_role`. The in-body JWT guard *does* close the realistic exposure (a `service_role` call through PostgREST carries `request.jwt.claims` and trips the 42501 guard), so this is not a live vulnerability. But for consistency with the stated intent ("granted only to postgres") and true defense-in-depth, add the missing revoke here (and mirror it in `03_functions.sql`):

```sql
REVOKE EXECUTE ON FUNCTION public.purge_orphaned_search_history() FROM service_role;
```

supabase/schemas/global_search/search_history_cleanup/README.md
```diff
@@ -0,0 +1,71 @@
...
+- Uses `NOT EXISTS` anti-joins against `auth.users(id)` (its PK). Note the plan is a seq scan + hash anti-join — every row must be checked, so the `user_id` indexes do **not** accelerate this query. That is fine at these tables' expected size (bounded by users × distinct terms); if they ever grow large, batch the DELETE.
+- **Not exposed via the API.** `EXECUTE` is revoked from `PUBLIC`, `anon`, and `authenticated`, and granted only to `postgres` (the role `pg_cron` runs as) — this REVOKE is the primary access control. As defense-in-depth against an *authenticated* call, the body also rejects any request carrying a non-empty `request.jwt.claims` JWT (which PostgREST sets for authenticated requests and `pg_cron` never does) with `42501`. This second layer matters because the CLI's `auto_expose_new_tables` grant pass can re-grant `EXECUTE` to the Data API roles on `db reset` (a repo-wide issue tracked by [CA-729](https://ripplearc.youtrack.cloud/issue/CA-729), affecting every function). Note the guard does not by itself stop an *anon* call (no JWT — indistinguishable from the cron path); that path relies on the REVOKE, as with every other `SECURITY DEFINER` function in this schema.
```
ripplearcgit: 🟡 **"granted only to `postgres`" wording**

This isn't literally accurate while `auto_expose_new_tables` is on — a `db reset` can re-grant EXECUTE to `service_role` (mitigated by the JWT guard, not by the grant list). Soften the wording to reflect that `service_role` may retain EXECUTE under auto-expose.

supabase/tests/functions/purge_orphaned_search_history_test.sql
```diff
@@ -0,0 +1,84 @@
+BEGIN;
+
+-- Tests for CA-597: public.purge_orphaned_search_history() deletes rows in
+-- search_history and project_search_history whose user_id no longer exists in
+-- auth.users, while leaving rows of active users untouched.
+
+SELECT plan(6);
```
ripplearcgit: 🟡 **Optional: assert the cron job registered**

The suite doesn't assert the `pg_cron` job actually registered (e.g. `SELECT ... FROM cron.job WHERE jobname = 'purge-orphaned-search-history'`). The cron wiring is the one thing `supabase db diff` can't verify, so a single existence check (bumping `plan(6)` to `plan(7)`) would be the highest-value addition — purely optional, since the function behaviour is already well covered.

ripplearcgit (approved):
## ⚠ Conditional LGTM

Approach is sound — `SECURITY DEFINER` is correctly justified, `search_path` is pinned to the repo convention, the JWT-guard logic is correct (blocks only non-empty claims), and the migration/schema split is handled cleanly with the cron caveat documented. No blockers; 3 polish/consistency nits posted inline that the author can resolve on merge.

### Architectural impact

```mermaid
sequenceDiagram
    participant Cron as pg_cron (postgres)
    participant API as PostgREST (anon/auth)
    participant Fn as purge_orphaned_search_history()
    participant SH as search_history
    participant PSH as project_search_history
    participant AU as auth.users
    Cron->>Fn: SELECT purge() — no jwt.claims
    Note over Fn: guard passes (claims empty)
    Fn->>AU: NOT EXISTS anti-join
    Fn->>SH: DELETE orphaned rows
    Fn->>PSH: DELETE orphaned rows
    Fn-->>Cron: void
    API->>Fn: rpc call — jwt.claims set
    Note over Fn: guard trips
    Fn-->>API: RAISE 42501
```

⚠ **Watch:** the REVOKE list omits `service_role`; with `auto_expose_new_tables = true` (CA-729) that role can be re-granted EXECUTE on `db reset`. The in-body guard closes the realistic (PostgREST) exposure, but the doc's "granted only to postgres" is not literally true. Non-blocking.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-backend-45.html

PR #46, opened 2026-07-24
CA-596: feat(search): apply filter_by_tag to global_search via project_tags

Bazabizi:
## Summary

Backend half of [CA-596](https://ripplearc.youtrack.cloud/issue/CA-596): the Flutter tag filter sheets already send `filter_by_tag` on every search, but `global_search` accepted the parameter without applying it — no project-to-tag association existed in the schema, so users selecting a tag got unfiltered results.

- **New `project_tags` pivot** (`schemas/core/project_tags/`): many-to-many `projects` ↔ `tags`, modeled on the existing `session_tags` pivot (`id`, `project_id` FK, `tag_id` FK, `applied_at`, `UNIQUE (project_id, tag_id)`, index on `tag_id`). Both FKs are `ON DELETE CASCADE` — a dangling link is meaningless on either side.
- **RLS**: SELECT follows project visibility via `user_has_project_permission(project_id, 'view_project', auth.uid())` — the same predicate as `projects_select_policy`, so a tag assignment is never visible (RPC or Data API) to someone who can't view the project. No write policies: writes are `service_role`/migrations only, matching the `tags` reference-data pattern.
- **Migration 39** (generated via `supabase db diff --from=migrations --to=local` from the declarative schema files, per the module README workflow): creates the table/indexes/policy and recreates `public.global_search` with the tag predicate applied to the projects block: `filter_by_tag IS NULL OR EXISTS (… project_tags JOIN tags ON tags.id = tag_id WHERE project_id = projects.id AND tags.name = filter_by_tag)`.
- **Contract**: `NULL` means "no tag filter" (behavior unchanged); the match is exact against `tags.name`; a name matching no tag returns zero projects. The filter applies **only to the projects result** — estimations and members are unaffected. The RPC signature is unchanged.
- **Schema backfill**: `global_search/global_search/03_functions.sql` mirrors the migration body exactly; both CA-596 TODO markers and the "reserved; not yet applied" notes are removed; module READMEs updated (`core/project_tags/` added, `global_search` parameter table updated).

## Test plan

- New `global_search_tag_filter_test.sql` (13 pgTAP assertions): NULL "no filter" contract, exact-name narrowing, a tag on two projects returning both, an unused tag returning zero, tag×date-range intersection, the estimations path being unaffected by `filter_by_tag`, the RLS boundary (a tagged project the viewer isn't a member of is excluded from search results *and* its pivot row is invisible to direct SELECT), both `ON DELETE CASCADE` sides, a `throws_ok` INSERT-as-authenticated write-denial pin, plus structural pins (`has_table`, unique-index shape). Runs under `SET LOCAL ROLE authenticated` with JWT `app_metadata.projects` claims for cost_estimates RLS.
- Verified locally with the CI-pinned Supabase CLI 2.106.0: `supabase db reset` applies migration 39 cleanly on a fresh DB; `supabase test db` passes all suites (185/185 across all 17 suites on this branch; 191 including the stacked #47).

### DIFF
--- a/supabase/migrations/20260724074549_39_global_search_tag_filter.sql
+++ b/supabase/migrations/20260724074549_39_global_search_tag_filter.sql
@@ -1,58 +1,20 @@
--- Functions for the Global Search feature (dashboard-wide search across
--- projects, cost estimates, and members).
+-- CA-596: Apply global_search's filter_by_tag param to the projects
+-- result. Introduces the project_tags pivot (projects <-> tags;
+-- previously no project-to-tag association existed) and filters the
+-- projects block via an EXISTS probe matching tags.name exactly.
+-- NULL filter_by_tag means "no tag filter" (existing behavior).
+-- SELECT on project_tags follows project visibility
+-- (user_has_project_permission), so tag assignments of invisible
+-- projects never surface; no write policies are defined (writes are
+-- service_role / migrations only, matching the tags pattern).
+-- Generated via `supabase db diff --from=migrations --to=local` from
+-- supabase/schemas/core/project_tags/ and
+-- supabase/schemas/global_search/global_search/03_functions.sql per
+-- that module's README workflow.
+-- https://ripplearc.youtrack.cloud/issue/CA-596
 
--- ============================================================
--- global_search RPC
---
--- Parameters:
---   query                 text     — required, matched against project_name/
---                                     description, estimate_name/description,
---                                     and member first_name/last_name/email.
---   filter_by_tag         text     — restricts the projects result to
---                                     projects linked (via project_tags) to
---                                     a tag with this exact name. NULL means
---                                     no tag filter. Applies only to
---                                     projects; estimations and members are
---                                     unaffected.
---   filter_by_date_from   timestamptz — inclusive lower bound on updated_at.
---   filter_by_date_to     timestamptz — inclusive upper bound on updated_at.
---   filter_by_owners      uuid[]   — restricts projects/estimates to rows
---                                     whose creator_user_id is in the array.
---                                     NULL or an empty array means "no owner
---                                     filter" (an empty array is the natural
---                                     "no owners selected" client state and
---                                     must not silently match zero rows).
---   scope                 text     — one of 'dashboard', 'estimation',
---                                     'member', or NULL for all.
---   projects_offset, estimations_offset, members_offset, "limit" — pagination.
---
--- Returns: jsonb { projects: [...], estimations: [...], members: [...] }
---
--- Date range filter:
---   Applies to projects.updated_at and cost_estimates.updated_at as an
---   inclusive range. Either bound may be NULL to leave that side
---   unbounded. An inverted range (filter_by_date_from > filter_by_date_to)
---   raises a 22023 error rather than silently returning no rows.
---
--- SECURITY INVOKER: relies on RLS on projects/cost_estimates/project_members/
--- users for row visibility — this function does not bypass RLS.
---
--- Privacy: the members result selects only u.id, first_name, last_name,
--- professional_role, profile_photo_url. Never select users.credential_id
--- here (prior review caught credential_id leakage on this exact RPC).
--- ============================================================
-CREATE OR REPLACE FUNCTION public.global_search(
-  query text,
-  filter_by_tag text DEFAULT NULL::text,
-  filter_by_date_from timestamp with time zone DEFAULT NULL::timestamp with time zone,
-  filter_by_date_to timestamp with time zone DEFAULT NULL::timestamp with time zone,
-  filter_by_owners uuid[] DEFAULT NULL::uuid[],
-  scope text DEFAULT NULL::text,
-  projects_offset integer DEFAULT 0,
-  estimations_offset integer DEFAULT 0,
-  members_offset integer DEFAULT 0,
-  "limit" integer DEFAULT 20
-)
+SET check_function_bodies = false;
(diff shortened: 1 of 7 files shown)


supabase/migrations/20260724074549_39_global_search_tag_filter.sql
```diff
@@ -0,0 +1,159 @@
...
+  );
+
+END;
+$function$;
+CREATE TABLE public.project_tags (id uuid DEFAULT gen_random_uuid() NOT NULL, project_id uuid NOT NULL, tag_id uuid NOT NULL, applied_at timestamp with time zone DEFAULT now() NOT NULL);
+ALTER TABLE public.project_tags ENABLE ROW LEVEL SECURITY;
+ALTER TABLE public.project_tags ADD CONSTRAINT project_tags_pkey PRIMARY KEY (id);
+ALTER TABLE public.project_tags ADD CONSTRAINT project_tags_project_id_fkey FOREIGN KEY (project_id) REFERENCES public.projects(id) ON DELETE CASCADE;
+ALTER TABLE public.project_tags ADD CONSTRAINT project_tags_tag_id_fkey FOREIGN KEY (tag_id) REFERENCES public.tags(id) ON DELETE CASCADE;
+GRANT ALL ON public.project_tags TO anon;
```
ripplearcgit, line 154: 🟡 **Nit (non-blocking, already mitigated): auto-exposed `GRANT ALL` includes `TRUNCATE`**

Table-level `GRANT ALL` to `anon`/`authenticated` includes `TRUNCATE`, which Postgres RLS does not gate at all (RLS only applies to row-returning DML, not `TRUNCATE`). This is Supabase's `auto_expose_new_tables` default re-applying on every `db diff`/`db reset` (tracked as CA-729 per the README), not something this PR introduced deliberately — and it's unreachable in practice since PostgREST's Data API never exposes a `TRUNCATE` operation to `anon`/`authenticated` JWT-scoped clients. Same shape as the `service_role` EXECUTE gap noted in the prior PR (#45) review — worth a standing follow-up ticket to decide whether to `REVOKE TRUNCATE` explicitly across all auto-exposed tables for defense-in-depth, rather than relying solely on "the API can't reach it."

Not a merge blocker.

supabase/schemas/global_search/global_search/03_functions.sql
```diff
@@ -99,6 +99,16 @@ BEGIN
...
         AND (filter_by_date_to IS NULL OR updated_at <= filter_by_date_to)
         AND (filter_by_owners IS NULL OR cardinality(filter_by_owners) = 0 OR creator_user_id = ANY(filter_by_owners))
+        AND (
+          filter_by_tag IS NULL
+          OR EXISTS (
+            SELECT 1
+            FROM project_tags pt
+            JOIN tags t ON t.id = pt.tag_id
+            WHERE pt.project_id = projects.id
+              AND t.name = filter_by_tag
```
ripplearcgit, line 109: 🟡 **Nit (optional, contract already documented): case-sensitive exact match on `tags.name`**

Every other text predicate in this function (project name/description, estimate name/description, member name/email) is wrapped in `lower(...)` for case-insensitive `LIKE`, but the tag join uses a bare `t.name = filter_by_tag`. This is a deliberate, documented contract ("the match is exact against tags.name") rather than a bug, and is almost certainly fine if the Flutter client always passes the canonical tag string from a picker rather than free-typed input — worth a quick confirmation on the client side, not a backend change.

ripplearcgit (approved):
### Architectural overview

```mermaid
sequenceDiagram
    participant Client as Flutter Client
    participant RPC as global_search SECURITY INVOKER
    participant Proj as projects table RLS
    participant PT as project_tags table RLS
    participant Tags as tags reference table

    rect rgba(139,92,246,0.15)
    Note over Client,Tags: Analogy: filtering a stack of already-readable papers. A filter can only remove papers - never add one you were not allowed to see.
    end

    Client->>RPC: global_search(query, filter_by_tag='concrete', scope='dashboard')
    RPC->>Proj: SELECT projects WHERE name/desc match
    Note over Proj: projects_select_policy already restricts rows to viewer's visible projects
    RPC->>PT: EXISTS (project_tags JOIN tags WHERE project_id=projects.id AND tags.name=filter_by_tag)
    Note over PT: project_tags_select_policy mirrors projects_select_policy (same predicate)
    alt viewer can view the project AND it carries the tag
        PT-->>RPC: EXISTS true
        RPC-->>Client: project included in results
        rect rgba(139,92,246,0.15)
        Note over RPC,Client: the paper stays in the stack
        end
    else viewer cannot view the project (not a member)
        Proj-->>RPC: row already excluded, EXISTS probe never reached
        RPC-->>Client: project absent (belt-and-suspenders double gate)
        rect rgba(139,92,246,0.15)
        Note over RPC,Client: the paper was never in the stack
        end
    end
```

Approach is sound: `global_search` stays `SECURITY INVOKER`, so the new tag `EXISTS` probe can only *narrow* a result set that `projects_select_policy` already restricted — never widen it. `project_tags_select_policy` mirrors `projects_select_policy` byte-for-byte, and the schema file matches the migration's function body exactly (future `db diff` sees no drift). pgTAP: 13/13 assertions, covering both the direct and indirect RLS boundary, write-denial, and both `ON DELETE CASCADE` directions.

### Nits (non-blocking)

- 🟡 `supabase/migrations/20260724074549_39_global_search_tag_filter.sql` — auto-exposed `GRANT ALL` includes `TRUNCATE` (unreachable via PostgREST; standing follow-up ticket suggested). See inline comment.
- 🟡 `supabase/schemas/global_search/global_search/03_functions.sql` — case-sensitive exact match on `tags.name` (documented contract, not a bug; worth confirming client always sends canonical tag values). See inline comment.

**Suggested next steps** (optional, non-blocking):
- Confirm the Flutter tag-filter sheet always sends canonical `tags.name` values (from a picker) rather than free-typed text.
- Consider a repo-wide ticket to explicitly `REVOKE TRUNCATE` on auto-exposed tables for defense-in-depth.
- Merge order: this PR must land before #47 (CA-839, stacked on this branch).

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-backend-46.html

PR #48, opened 2026-08-15
CA-963: feat(consent): add consent_versions and the current_consent_versions view

SabonaWaktole:
## Summary
- Adds `consent_versions` (publication history of each consent document) and `current_consent_versions` (highest live version per type, respecting `effective_from`)
- Read-only for clients via the view — SELECT-only RLS, no write policy for any role, so publishing stays a migration-only act
- Seeds version 1 of both consent types with provisional document URLs (must be confirmed before `CONSENT_GATE_ENABLED` goes live)

## Stack
1. **This PR** — `consent_versions` schema + seed
2. #next — `user_consents` append-only log (stacked on this branch)

## Test plan
- [x] `supabase/tests/database/consent_versions_test.sql` covers RLS read-only behavior, blank `document_url` rejection, and `current_consent_versions` selection logic
- [x] `npx supabase db reset` locally to confirm migration applies cleanly

### DIFF
--- a/supabase/migrations/20260814125300_42_seed_consent_versions_v1.sql
+++ b/supabase/migrations/20260814125300_42_seed_consent_versions_v1.sql
@@ -1,0 +1,23 @@
+-- CA-963: Publish version 1 of each consent document.
+--
+-- A migration rather than a seeder: supabase/seeders/ only runs on local
+-- `start` / `db reset`, so a seeded row would never reach production — and an
+-- empty consent_versions leaves every gate check resolving to
+-- ConsentIndeterminate, which blocks nothing but presents nothing either.
+--
+-- TODO: the document_url values below are PROVISIONAL. No published legal
+-- document URL existed when this was written; these mirror the shape the
+-- client's temporary in-memory source fakes. Confirm both against the real
+-- published documents before CONSENT_GATE_ENABLED is switched on in
+-- production. https://ripplearc.youtrack.cloud/issue/CA-963
+--
+-- ON CONFLICT DO NOTHING so a re-run against a database that already has v1
+-- (a partially applied deploy, a re-seeded branch) is a no-op rather than a
+-- unique violation that fails the whole migration.
+
+INSERT INTO public.consent_versions
+  (consent_type, version, document_url, effective_from, published_at)
+VALUES
+  ('terms_and_privacy', 1, 'https://ripplearc.com/legal/terms-and-privacy', now(), now()),
+  ('analytics',         1, 'https://ripplearc.com/legal/analytics',         now(), now())
+ON CONFLICT (consent_type, version) DO NOTHING;
--- a/supabase/schemas/_types/enums.sql
+++ b/supabase/schemas/_types/enums.sql
@@ -189,3 +189,15 @@ CREATE TYPE "public"."cost_estimation_activity_type_enum" AS ENUM (
 );
 
 ALTER TYPE "public"."cost_estimation_activity_type_enum" OWNER TO "postgres";
+
+
+-- Consent documents, versioned independently of one another. Values are the
+-- wire contract with the Flutter client (consent_wire_values.dart).
+
+CREATE TYPE "public"."consent_type_enum" AS ENUM (
+    'terms_and_privacy',
+    'analytics'
+);
+
+
+ALTER TYPE "public"."consent_type_enum" OWNER TO "postgres";
--- a/supabase/schemas/consent/consent_versions/03_rls.sql
+++ b/supabase/schemas/consent/consent_versions/03_rls.sql
@@ -1,0 +1,12 @@
+-- RLS policies for consent_versions
+-- Read-only for clients. No INSERT/UPDATE/DELETE policy for any role: the
+-- absence is the enforcement point, since auto-expose (CA-729) grants the Data
+-- API roles full SQL access. service_role bypasses RLS, which is how the seed
+-- migration writes. Reads are unscoped — the terms are the same document for
+-- everyone. See README.md.
+
+ALTER TABLE "public"."consent_versions" ENABLE ROW LEVEL SECURITY;
+
+CREATE POLICY "consent_versions_select_policy" ON "public"."consent_versions"
+  FOR SELECT TO "authenticated"
+  USING (true);
--- a/supabase/tests/database/consent_versions_test.sql
+++ b/supabase/tests/database/consent_versions_test.sql
@@ -1,0 +1,153 @@
+BEGIN;
+
+-- Tests for CA-963: consent_versions and the current_consent_versions view.
+-- Covers the enum wire contract with the Flutter client, the constraints that
+-- keep an unusable requirement out of the table, the "which version is in
+-- force" rules the view exists to own, the v1 seed, and the read-only-for-
+-- clients RLS posture.
+
+SELECT plan(14);
+
+-- ============================================================
+-- Shape
+-- ============================================================
+
+SELECT has_table('public', 'consent_versions', 'consent_versions table should exist');
+SELECT has_pk('public', 'consent_versions', 'consent_versions should have a primary key');
+
+-- Wire contract: these literals are compared against strings the Flutter
+-- client sends (consent_wire_values.dart). Renaming a value here silently
+-- stops matching on the client, so the exact set is pinned.
+SELECT is(
+  (SELECT array_to_string(array_agg(e.enumlabel ORDER BY e.enumsortorder), ',') COLLATE "C"
+     FROM pg_enum e
+     JOIN pg_type t ON t.oid = e.enumtypid
+     WHERE t.typname = 'consent_type_enum'),
+  'terms_and_privacy,analytics' COLLATE "C",
+  'consent_type_enum carries exactly the client wire values, in order'
+);
+
+SELECT ok(
+  EXISTS (
+    SELECT 1 FROM pg_constraint
+    WHERE conname = 'consent_versions_type_version_uq'
+      AND conrelid = 'public.consent_versions'::regclass
+      AND contype = 'u'
+  ),
+  'consent_versions has a UNIQUE constraint on (consent_type, version)'
+);
+
+-- ============================================================
+-- Constraints — an unusable requirement must not reach the table
+-- ============================================================
+
+SELECT throws_ok(
+  $$INSERT INTO public.consent_versions (consent_type, version, document_url)
+    VALUES ('analytics', 0, 'https://example.com/doc')$$,
+  '23514',
+  NULL,
+  'A version of 0 is rejected (versions are 1-based)'
+);
+
+-- A blank URL would present a consent page with nothing to read: the client
+-- throws on it and resolves the throw to a status rather than an error, so the
+-- failure would be invisible at runtime. Rejected on write instead.
+SELECT throws_ok(
+  $$INSERT INTO public.consent_versions (consent_type, version, document_url)
+    VALUES ('analytics', 99, '   ')$$,
+  '23514',
+  NULL,
+  'A blank document_url is rejected'
+);
+
+-- ============================================================
+-- Seed
+-- ============================================================
+
+SELECT is(
+  (SELECT count(*) FROM public.consent_versions WHERE version = 1),
+  2::bigint,
+  'The seed migration published version 1 of both consent types'
+);
+
+-- ============================================================
+-- current_consent_versions — the rules the view exists to own
+-- ============================================================
+
+DO $$
+BEGIN
+  -- v2 is live, v3 is scheduled for next week. Seeded v1 is already present.
+  INSERT INTO public.consent_versions (consent_type, version, document_url, effective_from)
+  VALUES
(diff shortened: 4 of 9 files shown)


supabase/schemas/consent/consent_versions/03_rls.sql
```diff
@@ -0,0 +1,12 @@
...
+-- Read-only for clients. No INSERT/UPDATE/DELETE policy for any role: the
+-- absence is the enforcement point, since auto-expose (CA-729) grants the Data
+-- API roles full SQL access. service_role bypasses RLS, which is how the seed
+-- migration writes. Reads are unscoped — the terms are the same document for
+-- everyone. See README.md.
+
+ALTER TABLE "public"."consent_versions" ENABLE ROW LEVEL SECURITY;
+
+CREATE POLICY "consent_versions_select_policy" ON "public"."consent_versions"
+  FOR SELECT TO "authenticated"
```
ripplearcgit, line 11: Nit — no `anon` read policy. Only `authenticated` gets a SELECT policy here; a pre-auth/logged-out caller resolves to `anon` and gets an empty result from `current_consent_versions`, not an error. Very likely intentional (consent gate presumably shown post-auth per CA-962), so this is a confirmation ask, not a change request — please confirm the flow timing matches this assumption.

supabase/migrations/20260814125300_42_seed_consent_versions_v1.sql
```diff
@@ -0,0 +1,23 @@
...
+-- production. https://ripplearc.youtrack.cloud/issue/CA-963
+--
+-- ON CONFLICT DO NOTHING so a re-run against a database that already has v1
+-- (a partially applied deploy, a re-seeded branch) is a no-op rather than a
+-- unique violation that fails the whole migration.
+
+INSERT INTO public.consent_versions
+  (consent_type, version, document_url, effective_from, published_at)
+VALUES
+  ('terms_and_privacy', 1, 'https://ripplearc.com/legal/terms-and-privacy', now(), now()),
```
ripplearcgit, line 21: Nit — already ticketed, flagging only so it isn't lost before go-live. These `document_url` values are explicitly provisional (TODO + CA-963 above). Flagging here only to make sure "confirm real URLs before `CONSENT_GATE_ENABLED` goes live" survives as a release-gate checklist item, not just a comment that scrolls out of view.

supabase/schemas/_types/enums.sql
```diff
@@ -189,3 +189,15 @@ CREATE TYPE "public"."cost_estimation_activity_type_enum" AS ENUM (
 );
 
 ALTER TYPE "public"."cost_estimation_activity_type_enum" OWNER TO "postgres";
+
+
+-- Consent documents, versioned independently of one another. Values are the
+-- wire contract with the Flutter client (consent_wire_values.dart).
+
+CREATE TYPE "public"."consent_type_enum" AS ENUM (
```
ripplearcgit, line 197: Nit — `consent_type_enum` is now declared in two hand-maintained places (here and `supabase/migrations/20260814125200_41_consent_versions.sql` lines 28-31), with no tooling enforcing they stay identical (a consequence of `schema_paths = []`, and a pre-existing repo convention this PR follows correctly rather than introduces). Non-blocking suggestion: a cheap CI check diffing the two `CREATE TYPE`/`CREATE TABLE` statement sets would catch future drift before it becomes a silent production bug.

ripplearcgit (approved):
## Verdict: CONDITIONAL LGTM (approve, non-blockers only)

Schema is sound and no blockers. The specific failure modes this review looks for on a schema PR — wrong column types, missing constraints, a "current version" view that can return stale/inconsistent data, an inappropriate `SECURITY DEFINER`, missing RLS, missing index coverage — are all either correctly handled or not applicable:

- The `UNIQUE (consent_type, version)` constraint on `consent_versions` makes `current_consent_versions`'s `DISTINCT ON (consent_type) … ORDER BY version DESC` deterministic by construction, not by convention — two rows can never tie for "current" within a `consent_type`.
- `SECURITY INVOKER` is the correct, explained choice — the view runs with the caller's own permissions, not an elevated one.
- RLS enables read access without opening any write path — the *absence* of an INSERT/UPDATE/DELETE policy is the actual enforcement, verified in pgTAP for both RLS failure shapes (`42501` on INSERT vs. silent zero-rows on UPDATE/DELETE).
- The composite `UNIQUE` index already serves the view's query pattern; no separate index is needed.
- 14-assertion pgTAP suite matches `plan(14)` and covers shape, both CHECK constraints, the seed, both view resolution rules, and all three write paths.

Two non-blocking nits below — both worth a confirming look, neither blocks merge.

### Nit 1 — no `anon` read policy (confirm this is intentional)

Think of RLS as a bouncer checking ID at the door on every read and write, not a single lock on the whole table. Right now the bouncer has exactly one rule on the guest list: "if you're `authenticated`, you may read." There is no rule at all for `anon` (a logged-out / pre-session caller) — so `anon` isn't blocked with an error, it just gets an empty result set back from `current_consent_versions`, which is a confusing thing to debug later if it ever happens by surprise.

This is very likely correct by design — the consent gate is presumably shown only after the JWT is issued — but it's worth an explicit confirmation with whoever owns the CA-962 flow timing, since nothing in this diff enforces that assumption.

`supabase/schemas/consent/consent_versions/03_rls.sql` line 11 (`FOR SELECT TO "authenticated"`).

### Nit 2 — consent-type enum is duplicated across two hand-maintained files, no CI check keeping them in sync

`consent_type_enum` is declared identically in two places — the migration and the declarative schema file — because `config.toml` sets `schema_paths = []`, so `supabase db diff` has nothing to diff against and the two trees have to be kept in step by hand. This PR keeps them correctly in sync, but nothing currently *catches* the two definitions drifting apart in a future PR — it's the enum equivalent of keeping two copies of the same list in two files with no test that fails if someone edits only one. A cheap CI check that diffs the `CREATE TYPE`/`CREATE TABLE` statement sets between `supabase/schemas/**` and the matching migration would close that gap. This is a pre-existing repo convention, not something introduced by this PR — flagging as a general, non-blocking suggestion.

`supabase/schemas/_types/enums.sql` line 197 (`CREATE TYPE "public"."consent_type_enum" AS ENUM (`), duplicated in `supabase/migrations/20260814125200_41_consent_versions.sql` lines 28–31.

Also flagged inline for visibility (already tracked, not a gap): the seeded `document_url` values in the v1 seed migration are explicitly provisional (TODO + CA-963) — call-out is only so it doesn't get lost before `CONSENT_GATE_ENABLED` goes live, not a request to change this diff.

<details>
<summary>Suggested next steps</summary>

- Confirm with the CA-962 design (or whoever owns the consent-gate flow timing) that the consent screen is always shown post-authentication, since only `authenticated` — not `anon` — has a SELECT policy on this table. If any pre-auth surface needs to read consent terms, that will need its own policy in a follow-up.
- Before `CONSENT_GATE_ENABLED` is switched on in production, replace the seeded provisional `document_url` values with the real published legal document URLs (already tracked as a TODO against CA-963 in this diff — just don't let it slip).
- Optional, low-priority, repo-wide (not blocking this PR): consider a CI check that diffs `supabase/schemas/**` against the equivalent hand-written migration DDL, so the two trees (kept in step by hand today, here and elsewhere in the repo) can't silently drift.
- When the follow-up PR adds `user_consents`, double check its `version` column actually uses `CHECK (version >= 0)` as this PR's README promises, not a copy-pasted `> 0`.

</details>

**Pre-merge checklist**
- No changes required inside this diff before merging
- Confirm the consent gate's pre-auth vs. post-auth timing against the `authenticated`-only RLS policy
- Track "replace provisional document_url seed values" as a release gate ahead of `CONSENT_GATE_ENABLED`, separate from this merge

supabase/schemas/consent/consent_versions/03_rls.sql
SabonaWaktole: Confirmed — the gate is post-authentication only, so the `authenticated`-only policy is correct and `anon` never needs to read this.

Traced it in `construculator-app`: every `ConsentRepositoryImpl` entry point takes `SupabaseWrapper.getInternalUserId()` before doing anything. The three read paths — `getCachedConsentStatus`, `watchConsentStatus`, `verifyPublishedVersion` — return `_ungated` (`ConsentSatisfied`) on `null`; the two write paths hand the nullable id to `ConsentRecorder`, which fails it as `ConsentErrorType.authenticationError`. `verifyPublishedVersion` is the only path that reaches `RemoteConsentDataSource`, and so `current_consent_versions`, and its null check sits above that call — an unauthenticated caller never issues the query at all.

No pre-auth surface reads the table either. The signup screen's terms links are `l10n.termsAndServicesLink` / `l10n.privacyPolicyLink` with a static `_openLink('')` handler — they don't come from `current_consent_versions`. `ConsentDocumentLinks` (the widget that does render published URLs) is used only by `ConsentPrompt`, i.e. the gate itself.

If a pre-auth surface ever needs the terms, it'll need its own policy — agreed that's a follow-up rather than something to add speculatively now.

supabase/tests/database/consent_versions_test.sql
```diff
@@ -0,0 +1,153 @@
...
+-- RLS — readable by any authenticated user, writable by none
+-- ============================================================
+
+SET LOCAL ROLE authenticated;
+SELECT set_config('request.jwt.claims', '{"sub": "22222222-2222-2222-2222-222222222222"}', true);
+
+SELECT ok(
+  (SELECT count(*) FROM public.consent_versions) > 0,
+  'An authenticated user can read published consent versions'
+);
```
ayanasamuel8, line 144: The view-resolution assertions (lines 86–106) run as the migration/superuser role, *before* `SET LOCAL ROLE authenticated` on line 112 — so they prove the `DISTINCT ON` logic but bypass RLS. The `authenticated` block that follows only ever selects the base table `consent_versions` (line 116), never `current_consent_versions`. Net result: nothing exercises the path the client actually uses — an `authenticated` caller reading `current_consent_versions` through `security_invoker` + the SELECT policy + the Data-API grant. A regression that flipped `security_invoker`, dropped the grant, or broke the policy would leave all 14 assertions green. Worth one assertion after line 118:

```sql
SELECT ok(
  (SELECT count(*) FROM public.current_consent_versions) > 0,
  'An authenticated user can read the resolver view (security_invoker composes with the SELECT policy)'
);
```

supabase/tests/database/consent_versions_test.sql
```diff
@@ -0,0 +1,153 @@
...
+
+-- The whole point of effective_from: a version can be inserted ahead of time
+-- without going live. If this fails, publishing v3 early re-gates every user.
+SELECT is(
+  (SELECT count(*) FROM public.current_consent_versions
+     WHERE consent_type = 'terms_and_privacy' AND version = 3),
+  0::bigint,
+  'A version whose effective_from is still in the future is not yet in force'
+);
```
ayanasamuel8, line 106: The "none in force" branch is a documented contract — the seed migration's own comment and the README both note an empty `consent_versions` for a type leaves the gate resolving to `ConsentIndeterminate` — but it's untested: the seed guarantees both enum values always have a v1 row, so the view's "consent_type absent entirely" case never executes here. Worth locking in, e.g. (before the role switch on line 112) delete the `analytics` rows within the transaction and assert `current_consent_versions` then returns exactly the `terms_and_privacy` row and no `analytics` row. Non-blocking.

ayanasamuel8 (approved):
Approve with conditions. The schema is genuinely well-built and I found no correctness defect: `UNIQUE (consent_type, version)` makes the view's `DISTINCT ON (consent_type) … ORDER BY version DESC` tie-free by construction; `security_invoker = true` is the correct, explained choice so the base-table SELECT policy governs view access; write protection is the deliberate absence of INSERT/UPDATE/DELETE policies (consistent with the repo's `tags` / `professional_roles` reference-data pattern); `timestamptz` throughout; `text` + non-blank `CHECK` over `varchar(n)`; UUID surrogate PK + natural unique key matching every other table here. Column set matches the CA-963 spec exactly. Naming/numbering (41, 42), `OWNER TO postgres`, the declarative-schema mirror, README and test-table registration all follow convention. pgTAP suite matches `plan(14)`.

Cross-PR with #49: checked and consistent — `user_consents` reuses the enum/`version` types, deliberately carries no FK back here, numbering `41 → 42 → 43` and the shared `_types/enums.sql` append are clean, and the `version >= 0` vs `> 0` divergence is called out in the README.

Conditions:
- Two small test-composition gaps (inline) — the client's actual read path through the view isn't exercised, and the "consent type entirely absent" view branch isn't locked in.
- **CA-962 sign-off (process gate, not a code defect).** CA-963's description makes this a merge precondition: the design doc's Approver status is "Not started" and two review comments are unresolved, one of which — analytics as an independently-withdrawable consent type — determines the `consent_type_enum` values this PR bakes in and seeds. An enum value that has shipped to production is expensive to walk back. Please confirm on the PR that the CA-962 analytics comment is resolved and the doc has approver sign-off, or hold merge until both land. (This applies to the stack; #48 is the one targeting `master`.)

PR #49, opened 2026-08-15
CA-963: feat(consent): add the append-only user_consents log

SabonaWaktole:
## Summary
- Adds `user_consents`, an append-only log of every consent decision (acceptances and withdrawals) keyed by `jwt_internal_user_id()` rather than `auth.uid()`, since `auth.uid()` resolves to `credential_id`, not `users.id`
- No UPDATE/DELETE policy for anyone, including the row owner — that's what makes the log append-only
- Deliberately no FK to `consent_versions` and no uniqueness on `(user_id, consent_type, version)` — see commit message for rationale

## Stack
- Based on #48 (`consent_versions` schema) — merge that first, then rebase this onto `master`

## Test plan
- [x] `supabase/tests/database/user_consents_test.sql` covers RLS append-only semantics (blocked UPDATE/DELETE return zero rows, not errors) and withdrawal-without-prior-acceptance (`version = 0`)
- [x] `npx supabase db reset` locally to confirm migration applies cleanly on top of #48

### DIFF
--- a/supabase/migrations/20260814190000_43_user_consents.sql
+++ b/supabase/migrations/20260814190000_43_user_consents.sql
@@ -1,0 +1,64 @@
+-- CA-963: Add user_consents — the append-only log of consent decisions,
+-- withdrawals as well as acceptances. Completes the consent schema begun in
+-- migration 41 (consent_versions).
+--
+-- Adds consent_action_enum and the jwt_internal_user_id() RLS helper, which
+-- exists because auth.uid() returns users.credential_id while user_consents.
+-- user_id references users.id — a policy on auth.uid() matches nothing and the
+-- symptom looks like broken sync rather than a bug.
+--
+-- Written by hand rather than via `supabase db diff`: config.toml declares
+-- schema_paths = [], so the CLI has no declared schema to diff against.
+-- Mirrors supabase/schemas/consent/user_consents/ and the shared helper in
+-- supabase/schemas/_shared/01_functions.sql — keep them in step.
+-- https://ripplearc.youtrack.cloud/issue/CA-963
+
+CREATE TYPE public.consent_action_enum AS ENUM (
+  'accepted',
+  'withdrawn'
+);
+
+ALTER TYPE public.consent_action_enum OWNER TO postgres;
+
+CREATE OR REPLACE FUNCTION public.jwt_internal_user_id()
+RETURNS uuid
+LANGUAGE sql
+SECURITY INVOKER
+STABLE
+AS $$
+  SELECT NULLIF(auth.jwt() -> 'app_metadata' ->> 'internal_user_id', '')::uuid
+$$;
+
+ALTER FUNCTION public.jwt_internal_user_id() OWNER TO postgres;
+
+COMMENT ON FUNCTION public.jwt_internal_user_id() IS 'Shared RLS helper. Returns the caller''s public.users.id from the JWT app_metadata.internal_user_id claim, or NULL when the claim is absent. Use instead of auth.uid() on tables keyed by users.id -- auth.uid() resolves to users.credential_id and never matches. NULL never equals a NOT NULL user_id, so an absent claim denies rather than leaks.';
+
+CREATE TABLE IF NOT EXISTS public.user_consents (
+  id           uuid PRIMARY KEY DEFAULT gen_random_uuid(),
+  user_id      uuid NOT NULL REFERENCES public.users(id) ON DELETE CASCADE,
+  consent_type public.consent_type_enum NOT NULL,
+  version      integer NOT NULL,
+  action       public.consent_action_enum NOT NULL,
+  recorded_at  timestamptz NOT NULL DEFAULT now(),
+  app_version  text,
+  platform     text,
+  -- >= 0, not > 0: a withdrawal with nothing on file to revoke records 0.
+  CONSTRAINT user_consents_version_non_negative CHECK (version >= 0)
+);
+
+ALTER TABLE public.user_consents OWNER TO postgres;
+
+CREATE INDEX IF NOT EXISTS user_consents_user_type_recorded_idx
+  ON public.user_consents (user_id, consent_type, recorded_at DESC);
+
+ALTER TABLE public.user_consents ENABLE ROW LEVEL SECURITY;
+
+CREATE POLICY user_consents_select_policy ON public.user_consents
+  FOR SELECT TO authenticated
+  USING (user_id = public.jwt_internal_user_id());
+
+CREATE POLICY user_consents_insert_policy ON public.user_consents
+  FOR INSERT TO authenticated
+  WITH CHECK (user_id = public.jwt_internal_user_id());
+
+-- No UPDATE or DELETE policy, deliberately: the log is append-only.
--- a/supabase/schemas/consent/user_consents/01_table.sql
+++ b/supabase/schemas/consent/user_consents/01_table.sql
@@ -1,0 +1,34 @@
+-- user_consents table
+-- Append-only log of every consent decision — withdrawals as well as
+-- acceptances. Only the newest row per (user_id, consent_type) is meaningful
+-- to the gate; everything beneath it is history. No IP or user agent: CA-962
+-- specifies app_version and platform as the only context captured.
+-- See README.md.
+
+CREATE TABLE IF NOT EXISTS "public"."user_consents" (
+  "id"           uuid PRIMARY KEY DEFAULT (gen_random_uuid()),
+  "user_id"      uuid NOT NULL
+                   REFERENCES "public"."users"("id") ON DELETE CASCADE,
+  "consent_type" "public"."consent_type_enum" NOT NULL,
+  "version"      integer NOT NULL,
+  "action"       "public"."consent_action_enum" NOT NULL,
+  "recorded_at"  timestamptz NOT NULL DEFAULT (now()),
+  "app_version"  text,
+  "platform"     text,
+
+  -- NOT `> 0`, unlike consent_versions.version. A withdrawal records the
+  -- version it revokes, and the client writes 0 when there is nothing on file
+  -- to revoke; `> 0` would reject exactly those withdrawals at runtime.
+  CONSTRAINT "user_consents_version_non_negative" CHECK ("version" >= 0)
+);
+
+ALTER TABLE "public"."user_consents" OWNER TO "postgres";
+
+-- Two constraints deliberately ABSENT, both of which look like reasonable
+-- review suggestions:
+--   * No FK to consent_versions — the log must survive administrative
+--     correction of that table; a consent record is evidence and must not
+--     cascade away because someone fixed a document URL.
+--   * No UNIQUE on (user_id, consent_type, version) — accept, withdraw, then
+--     accept the same version again is legitimate, and all three are real
+--     events.
--- a/supabase/schemas/consent/user_consents/03_rls.sql
+++ b/supabase/schemas/consent/user_consents/03_rls.sql
@@ -1,0 +1,20 @@
+-- RLS policies for user_consents
+-- Personal ownership on read and insert; no update or delete path for anyone,
+-- which is what makes the table append-only. Withdrawal is an INSERT of a
+-- 'withdrawn' row, never an UPDATE of the acceptance it supersedes.
+--
+-- Keyed on jwt_internal_user_id(), NOT auth.uid(): auth.uid() returns
+-- users.credential_id while user_id here references users.id, so auth.uid()
+-- would match nothing and read as "consent isn't syncing" rather than as a bug.
+-- See README.md.
+
+ALTER TABLE "public"."user_consents" ENABLE ROW LEVEL SECURITY;
+
+CREATE POLICY "user_consents_select_policy" ON "public"."user_consents"
+  FOR SELECT TO "authenticated"
+  USING ("user_id" = "public"."jwt_internal_user_id"());
+
+-- WITH CHECK is what stops a user recording consent in someone else's name.
+CREATE POLICY "user_consents_insert_policy" ON "public"."user_consents"
+  FOR INSERT TO "authenticated"
+  WITH CHECK ("user_id" = "public"."jwt_internal_user_id"());
--- a/supabase/tests/database/user_consents_test.sql
+++ b/supabase/tests/database/user_consents_test.sql
@@ -1,0 +1,152 @@
+BEGIN;
+
+-- Tests for CA-963: user_consents, the append-only log of consent decisions.
+-- Covers the action enum wire contract, the constraints that must NOT exist
+-- (an FK to consent_versions, a uniqueness rule on re-acceptance), the
+-- version >= 0 allowance that withdrawals depend on, and the RLS posture:
+-- own rows only, keyed on the internal_user_id JWT claim rather than
+-- auth.uid(), with no update or delete path for anyone.
+
+SELECT plan(13);
+
+-- ============================================================
+-- Shape
+-- ============================================================
+
+SELECT has_table('public', 'user_consents', 'user_consents table should exist');
+SELECT has_pk('public', 'user_consents', 'user_consents should have a primary key');
+SELECT col_is_fk('public', 'user_consents', 'user_id', 'user_consents.user_id is a FK to users');
+
+SELECT is(
+  (SELECT array_to_string(array_agg(e.enumlabel ORDER BY e.enumsortorder), ',') COLLATE "C"
+     FROM pg_enum e
+     JOIN pg_type t ON t.oid = e.enumtypid
+     WHERE t.typname = 'consent_action_enum'),
+  'accepted,withdrawn' COLLATE "C",
+  'consent_action_enum carries exactly the client wire values, in order'
+);
+
+SELECT has_index(
+  'public', 'user_consents', 'user_consents_user_type_recorded_idx',
+  'The (user_id, consent_type, recorded_at) lookup index exists'
+);
+
+-- ============================================================
+-- Constraints that must NOT exist
+--
+-- Both would look like reasonable review additions, and both would break a
+-- real path — so their absence is asserted rather than left to comments.
+-- ============================================================
+
+SELECT is(
+  (SELECT count(*) FROM pg_constraint
+     WHERE conrelid = 'public.user_consents'::regclass
+       AND contype = 'f'
+       AND confrelid = 'public.consent_versions'::regclass),
+  0::bigint,
+  'user_consents has NO foreign key to consent_versions (log survives corrections there)'
+);
+
+SELECT is(
+  (SELECT count(*) FROM pg_constraint
+     WHERE conrelid = 'public.user_consents'::regclass
+       AND contype = 'u'),
+  0::bigint,
+  'user_consents has NO unique constraint (re-accepting after a withdrawal is legitimate)'
+);
+
+-- ============================================================
+-- Fixtures
+-- ============================================================
+
+DO $$
+DECLARE
+  v_role_id uuid := '55555555-5555-5555-5555-555555555555';
+BEGIN
+  INSERT INTO professional_roles (id, name) VALUES (v_role_id, 'Consent Test Role');
+
+  INSERT INTO users (id, credential_id, email, first_name, last_name, professional_role, created_at, user_status, user_preferences, country_code)
+  VALUES
+    ('11111111-1111-1111-1111-111111111111', '22222222-2222-2222-2222-222222222222',
+     'consent_owner@example.com', 'Consent', 'Owner', v_role_id, now(), 'active', '{}', '+1'),
+    ('33333333-3333-3333-3333-333333333333', '44444444-4444-4444-4444-444444444444',
+     'consent_other@example.com', 'Consent', 'Other', v_role_id, now(), 'active', '{}', '+1');
+
+  -- The other user's record: must never be visible to the owner below.
+  INSERT INTO user_consents (user_id, consent_type, version, action)
+  VALUES ('33333333-3333-3333-3333-333333333333', 'terms_and_privacy', 1, 'accepted');
+END $$;
+
+-- A full accept -> withdraw -> re-accept cycle on ONE version. This is the
+-- sequence a UNIQUE (user_id, consent_type, version) would break, and the
+-- withdrawal-with-nothing-on-file case is what version >= 0 exists for.
+SELECT lives_ok(
+  $$INSERT INTO public.user_consents (user_id, consent_type, version, action) VALUES
+      ('11111111-1111-1111-1111-111111111111', 'terms_and_privacy', 0, 'withdrawn'),
+      ('11111111-1111-1111-1111-111111111111', 'terms_and_privacy', 2, 'accepted'),
+      ('11111111-1111-1111-1111-111111111111', 'terms_and_privacy', 2, 'withdrawn'),
+      ('11111111-1111-1111-1111-111111111111', 'terms_and_privacy', 2, 'accepted')$$,
+  'A withdrawal at version 0 and a re-accept of the same version both insert cleanly'
+);
+
+SELECT throws_ok(
+  $$INSERT INTO public.user_consents (user_id, consent_type, version, action)
+    VALUES ('11111111-1111-1111-1111-111111111111', 'analytics', -1, 'accepted')$$,
+  '23514',
+  NULL,
+  'A negative version is still rejected'
+);
+
+-- ============================================================
+-- RLS — own rows only, via the internal_user_id claim
+-- ============================================================
+
+SET LOCAL ROLE authenticated;
+
+-- Note the claim shape: app_metadata.internal_user_id is users.id, NOT the
+-- sub/credential_id. A policy written against auth.uid() would compare
+-- credential_id to user_id and match nothing.
+SELECT set_config('request.jwt.claims', '{
+  "sub": "22222222-2222-2222-2222-222222222222",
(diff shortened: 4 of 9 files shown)


supabase/migrations/20260814190000_43_user_consents.sql
```diff
@@ -0,0 +1,64 @@
...
+  SELECT NULLIF(auth.jwt() -> 'app_metadata' ->> 'internal_user_id', '')::uuid
+$$;
+
+ALTER FUNCTION public.jwt_internal_user_id() OWNER TO postgres;
+
+COMMENT ON FUNCTION public.jwt_internal_user_id() IS 'Shared RLS helper. Returns the caller''s public.users.id from the JWT app_metadata.internal_user_id claim, or NULL when the claim is absent. Use instead of auth.uid() on tables keyed by users.id -- auth.uid() resolves to users.credential_id and never matches. NULL never equals a NOT NULL user_id, so an absent claim denies rather than leaks.';
+
+CREATE TABLE IF NOT EXISTS public.user_consents (
+  id           uuid PRIMARY KEY DEFAULT gen_random_uuid(),
+  user_id      uuid NOT NULL REFERENCES public.users(id) ON DELETE CASCADE,
```
ripplearcgit: **Blocking: `ON DELETE CASCADE` silently erases a user's consent history.**

`user_id uuid NOT NULL REFERENCES public.users(id) ON DELETE CASCADE` means that if the referenced `users` row is ever hard-deleted, every consent row for that user disappears with it — no error, no trace. For a table whose whole purpose is being permanent proof of what a user agreed to and when, that's the exact failure mode this table can't afford.

Every other FK to `users(id)` in this codebase (`cost_estimates.creator_user_id`/`locked_by_user_id`, `cost_files.uploaded_by_user_id`, `cost_estimate_logs.user_id`) uses the safer default — no `ON DELETE` clause, so the delete is simply blocked while referencing rows exist. This table is the only one that cascades.

This is also the same reasoning this PR's own README applies (correctly) to the FK it deliberately *doesn't* have (`consent_versions`): a consent record is evidence and shouldn't cascade away. That reasoning was never applied to this FK.

Not exploitable today (RLS blocks ordinary users, and no hard-delete path exists in this repo yet), but this is foundational schema under a 21-PR consent chain — worth fixing now rather than as an `ALTER TABLE` against live compliance data later.

**Suggested fix:** drop `ON DELETE CASCADE` to match the rest of the codebase's convention, or, if cascading is genuinely intended, document why with the same rigor given to the FK that's absent.

supabase/schemas/consent/user_consents/01_table.sql
```diff
@@ -0,0 +1,34 @@
...
+-- Append-only log of every consent decision — withdrawals as well as
+-- acceptances. Only the newest row per (user_id, consent_type) is meaningful
+-- to the gate; everything beneath it is history. No IP or user agent: CA-962
+-- specifies app_version and platform as the only context captured.
+-- See README.md.
+
+CREATE TABLE IF NOT EXISTS "public"."user_consents" (
+  "id"           uuid PRIMARY KEY DEFAULT (gen_random_uuid()),
+  "user_id"      uuid NOT NULL
+                   REFERENCES "public"."users"("id") ON DELETE CASCADE,
```
ripplearcgit: **Blocking: same issue as the migration file** (`supabase/migrations/20260814190000_43_user_consents.sql` line 38) — `ON DELETE CASCADE` on this FK silently erases a user's entire consent history if their `users` row is ever hard-deleted, with no error and no trace.

Every other FK to `users(id)` in this codebase uses the safer default (no `ON DELETE` clause, delete blocked while dependents exist). This table is the outlier, and it's the opposite of the cascade-avoidance reasoning this same file gives two lines below for the (deliberately omitted) FK to `consent_versions`.

Same fix as the migration file: drop `ON DELETE CASCADE` here too (schema file and migration must stay in sync), or explicitly document why cascading is intended.

ripplearcgit (changes requested):
## Verdict: Changes Requested

This is a well-built append-only log — the core guarantee holds up under scrutiny (see notes below) — but there's one blocking issue in the foreign key from `user_consents.user_id` to `users(id)`.

### The blocking issue, in plain terms

`user_consents` exists to be permanent proof of what a user agreed to and when — a compliance/audit trail. Right now its link back to the `users` table is written as:

```sql
user_id uuid NOT NULL REFERENCES public.users(id) ON DELETE CASCADE
```

`ON DELETE CASCADE` means: if the referenced `users` row is ever deleted, silently delete every row here that points to it too. For a table whose entire job is being permanent evidence, that's the one failure mode the design can't afford — a hard delete of a user (an admin action, a future GDPR-erasure flow, or a script bug run with elevated credentials that bypass RLS) would erase that user's entire consent history with no error and no trace.

This isn't a hunch — I independently verified the raw SQL in both places it's declared:
- `supabase/migrations/20260814190000_43_user_consents.sql` line 38
- `supabase/schemas/consent/user_consents/01_table.sql` lines 10–11

Both say the same thing: `ON DELETE CASCADE`.

### Why this is the outlier, not the norm

Every other foreign key to `users(id)` in this codebase uses the safer default (no `ON DELETE` clause at all, which simply blocks the delete while referencing rows exist, forcing someone to make an explicit decision first):

- `cost_estimates.creator_user_id` / `cost_estimates.locked_by_user_id`
- `cost_files.uploaded_by_user_id`
- `cost_estimate_logs.user_id`

`user_consents` is the only table in this PR — and the only `users(id)` FK anywhere in the repo — that cascades.

What makes this stand out even more: this same PR's README carefully argues *against* cascading for a different relationship — the (deliberately omitted) FK to `consent_versions` — reasoning that "a consent record is evidence and must not cascade away because someone fixed a document URL." That reasoning is exactly as true for the `users` FK, but it's never applied there. `README.md` lines 21 and 102 (`supabase/schemas/consent/user_consents/README.md`) state "FK to `users.id`, `ON DELETE CASCADE`" and "user deletion cascades here" as plain facts, with no discussion of whether that's the right call — the one case where this same logic was never applied.

To be fair both ways: this is latent, not exploitable today. RLS already blocks an ordinary authenticated user from deleting anyone's `users` row, and this repo's own `users/README.md` recommends soft-delete via `user_status` rather than hard deletion — no hard-delete code path exists yet. But this is foundational schema underneath a 21-PR consent chain in `construculator-app`; fixing the FK now is a one-line change, whereas fixing it after real consent rows accumulate means an `ALTER TABLE` against live compliance data.

**Recommended fix:** drop `ON DELETE CASCADE` on both declarations, matching the rest of the codebase's convention. If cascading truly is intended for a specific reason, that reason should get the same written-out justification the README gives the FK that's absent — not be left as an unremarked default.

### What's already right (verified, not just read)

- The append-only guarantee is real: RLS is enabled, exactly two policies exist (SELECT, INSERT, both scoped to `authenticated` and to `user_id = jwt_internal_user_id()`), and there is genuinely no UPDATE or DELETE policy for any role — including the row's owner. Checked this against `config.toml`'s `auto_expose_new_tables = true`: since RLS is on and no UPDATE/DELETE policy exists, Postgres denies those commands for every row regardless of the underlying grant — the missing policy, not a missing grant, is what's actually stopping tampering.
- `jwt_internal_user_id()` was checked against the actual `custom_access_token_hook` (not just trusted from its comment) — the claim shape genuinely matches.
- The two constraints called out in the README as "deliberately absent" — no FK to `consent_versions`, no UNIQUE on `(user_id, consent_type, version)` — are correctly justified and regression-tested. These are **not** defects; please don't re-add either.

supabase/schemas/consent/user_consents/03_rls.sql
```diff
@@ -0,0 +1,20 @@
...
+ALTER TABLE "public"."user_consents" ENABLE ROW LEVEL SECURITY;
+
+CREATE POLICY "user_consents_select_policy" ON "public"."user_consents"
+  FOR SELECT TO "authenticated"
+  USING ("user_id" = "public"."jwt_internal_user_id"());
+
+-- WITH CHECK is what stops a user recording consent in someone else's name.
+CREATE POLICY "user_consents_insert_policy" ON "public"."user_consents"
+  FOR INSERT TO "authenticated"
+  WITH CHECK ("user_id" = "public"."jwt_internal_user_id"());
```
ripplearcgit, line 20: **Nit:** no test exercises an `authenticated` caller whose JWT is missing the `internal_user_id` claim entirely (as opposed to "claim present, wrong value"). The hook populating this claim does a live lookup by `credential_id` and can return `NULL` during a first-login race before the `users` row exists. Per this file's own `NULLIF`/`NULL ≠ NOT NULL` reasoning, SELECT should return zero rows and INSERT should be rejected in that case — worth a `throws_ok`/`is()` pair pinning it, since that reasoning is currently untested.

supabase/tests/database/user_consents_test.sql
```diff
@@ -0,0 +1,152 @@
...
+
+SELECT plan(13);
+
+-- ============================================================
+-- Shape
+-- ============================================================
+
+SELECT has_table('public', 'user_consents', 'user_consents table should exist');
+SELECT has_pk('public', 'user_consents', 'user_consents should have a primary key');
+SELECT col_is_fk('public', 'user_consents', 'user_id', 'user_consents.user_id is a FK to users');
```
ripplearcgit, line 20: **Nit:** `col_is_fk` confirms `user_id` is *a* foreign key to `users`, but doesn't assert its `ON DELETE` behavior. Once the CASCADE finding (see the migration and `01_table.sql` files) is resolved — cascade kept and justified, or dropped to match repo convention — consider pinning that choice the same way the FK-to-`consent_versions` and UNIQUE constraints are pinned above, so a future edit to this FK shows up as a failing test instead of silent schema drift.

ripplearcgit (commented):
## Nits (non-blocking)

Two small gaps, neither blocking merge:

1. **No test for a missing `internal_user_id` claim.** The hook that populates `app_metadata.internal_user_id` does a live lookup by `credential_id` and can return `NULL` if the `users` row hasn't been created yet (e.g. a race during first login). Current tests only cover "claim present, matches own row" and "claim present, targets someone else's row" — not "claim entirely absent." Per the `NULLIF`/`NULL ≠ NOT NULL` reasoning already in the README, SELECT should return zero rows and INSERT should be rejected in that case — but that's currently untested. See inline comment on `03_rls.sql`.
2. **The FK's `ON DELETE` behavior isn't pinned by a test.** `col_is_fk` (test file line 18) confirms `user_id` is *a* FK to `users`, but not its delete behavior. Once the blocking CASCADE finding is resolved, worth adding a regression assertion the same way the two "deliberately absent" constraints are already pinned — so any future edit to this FK fails CI instead of drifting silently. See inline comment on the test file.

## For the record: things checked and found correct, not flagged

- **No FK from `user_consents` to `consent_versions`.** Reasoned through and regression-tested: the log must survive administrative correction of the versions table.
- **No `UNIQUE (user_id, consent_type, version)`.** Also reasoned through and regression-tested: accept → withdraw → re-accept of the same version is a real, legitimate sequence, and the test file exercises it directly.

Both of these look like reasonable review suggestions at first glance; they are not defects, and I'm not asking for either to be added.

## Summary

Outside the CASCADE finding (posted as blocking inline comments on the migration and schema-declaration files), this diff is solid: the append-only guarantee is real and correctly enforced through RLS's default-deny behavior, the RLS helper function was verified against the actual token-issuing hook rather than trusted from its comment, and the test suite (`plan(13)`, hand-verified) correctly asserts the trickier Postgres behaviors — like a policy-less UPDATE/DELETE succeeding silently with zero rows changed, rather than expecting an error that Postgres never raises.

**Pre-merge checklist:**
- [ ] Drop (or explicitly justify) `ON DELETE CASCADE` on `user_consents.user_id` in both the migration and the schema file, and update the module README (lines 21, 102) to match
- [ ] Everything else in this diff — the enum, the RLS helper, the two RLS policies, the index, the test suite — is ready to merge as-is

supabase/migrations/20260814190000_43_user_consents.sql
SabonaWaktole: Fixed -- dropped `ON DELETE CASCADE`, bare FK now blocks the delete instead. Regression-tested (confdeltype + an actual refused DELETE) in `f6cefba`.

supabase/schemas/consent/user_consents/01_table.sql
SabonaWaktole: Fixed in the same commit (`f6cefba`) -- kept in step with the migration.

supabase/schemas/consent/user_consents/03_rls.sql
SabonaWaktole: Fixed -- added the claim-absent case: a caller with no `internal_user_id` claim sees zero rows on SELECT and is rejected on INSERT. Covered in `f6cefba`.

supabase/tests/database/user_consents_test.sql
SabonaWaktole: Fixed -- added a `confdeltype` assertion (metadata) plus an actual refused DELETE (behavior), so either drifting back fails CI. Covered in `f6cefba`.

SabonaWaktole:
@ripplearcgit Already addressed, pushed as `f6cefba`:

**Blocking — `ON DELETE CASCADE`:** Dropped in both declarations (migration + schema file), matching every other `users(id)` FK in the repo. README updated (lines 21, 102) to stop claiming the cascade. Regression-tested with both a `confdeltype` catalog assertion and an actual refused `DELETE`.

**Nit 1 — missing-claim case:** Added coverage for an `authenticated` caller with no `internal_user_id` claim at all — SELECT returns zero rows, INSERT is rejected.

**Nit 2 — FK delete-behavior pinned:** Added the `confdeltype` assertion alongside the existing `col_is_fk` check.

Verified locally: `npx supabase db reset` applies cleanly, and `npx supabase test db` is green — 223/223 (`user_consents_test.sql` now at 17 assertions, up from 13).

Could you re-review?

ripplearcgit (approved):
## Re-review (round 2) — commit `5713729` → `f6cefba3`

This is a follow-up to my earlier `CHANGES_REQUESTED` review at `5713729`, which flagged one blocker
and two nits. All three are resolved at current HEAD. 0 new findings, 0 regressions.

### Architectural overview

Unchanged from round 1 — the append-only mechanism itself was never the issue; the gap was one FK's
delete behavior. The core enforcement flow is still accurate:

```mermaid
sequenceDiagram
    participant App as ConsentRepositoryImpl (client)
    participant PG as Postgres role: authenticated
    participant JWT as jwt_internal_user_id()
    participant Tbl as public.user_consents
    App->>PG: INSERT user_id, consent_type, version, action='accepted'
    PG->>JWT: evaluate WITH CHECK on INSERT policy
    JWT-->>PG: auth.jwt() app_metadata.internal_user_id
    alt user_id matches caller's internal_user_id
        PG->>Tbl: row appended, RLS check passes
        Tbl-->>App: insert succeeds
    else user_id belongs to someone else
        PG--xApp: 42501 insert violates row-level security policy
    end
    Note over App,Tbl: Withdrawal is a second INSERT (action='withdrawn'), never an UPDATE
    App->>PG: UPDATE user_consents SET action='accepted' WHERE ...
    PG->>Tbl: no UPDATE policy exists for any role
    Tbl-->>PG: zero rows match, statement succeeds
    PG-->>App: 0 rows affected -- history unchanged
```

### What was fixed since round 1

**🔴→✅ Blocker — `ON DELETE CASCADE` on the `user_consents.user_id` FK to `users(id)`.**
This was the only cascading `users(id)` FK anywhere in the repo, which meant a hard `DELETE` on a
`users` row would silently wipe that user's entire consent history — the one thing an append-only
evidence log cannot allow.

Fixed by dropping the cascade in both `supabase/migrations/20260814190000_43_user_consents.sql` and
its declarative-schema mirror `supabase/schemas/consent/user_consents/01_table.sql`, so the FK now
falls back to the implicit default (`NO ACTION`) — matching every other `users(id)` FK in the codebase.
Confirmed by reading the live file content on branch HEAD `f6cefba`, not just trusting the diff: the
clause is genuinely absent, not moved or renamed.

This is **not just a metadata fix** — it's backed by two new pgTAP assertions in
`supabase/tests/database/user_consents_test.sql`:
1. A catalog-level check that `pg_constraint.confdeltype = 'a'` (NO ACTION) for the FK — pins what the
   FK was *declared* to do.
2. A `throws_ok(...)` test that actually issues `DELETE FROM public.users WHERE id = ...` against a
   user with existing consent rows and asserts it raises Postgres error `23503` (foreign key
   violation) — pins what the FK *actually does*. This second test is what would catch drift that
   metadata alone would miss (e.g. a future `BEFORE DELETE` trigger on `users` that clears
   `user_consents` some other way while leaving the FK declaration untouched).

The README (`supabase/schemas/consent/user_consents/README.md`) was also updated so this omission now
gets the same evidentiary argument the other two "deliberately absent" constraints already had, and it
adds a point beyond the original ask: there's no user-erasure path anywhere in this repo today
(`user_status` is only active/inactive; nothing deletes a `users` row), so a future GDPR-style erasure
flow will have to make an explicit decision about this log rather than silently inherit whatever the
FK's default happened to be.

> **In plain English:** think of a consent record as a visitor badge on file at the front desk.
> `ON DELETE CASCADE` used to mean the badge got shredded automatically the instant its owner walked
> out the door for good. The fix removes that automatic shredding — now if someone tries to have the
> owner walk out while their badge is still on file, the front desk simply refuses, until someone
> deals with the badge on purpose. Concretely: "the badge" is a `user_consents` row, "walking out the
> door for good" is a hard `DELETE FROM users`, and "the front desk refuses" is the FK raising Postgres
> error `23503` now that the cascade is gone.

**🟡→✅ Nit — no test for a caller with the `internal_user_id` claim absent entirely.**
`custom_access_token_hook` can return `NULL` for `internal_user_id` on a first-login race (the `users`
row doesn't exist yet), which is a real, reachable state distinct from "claim present but empty." A new
test section sets a JWT with a `sub` but no `app_metadata.internal_user_id` key at all, and asserts:
`SELECT` returns 0 rows, and `INSERT` throws `42501`. This is the correct "claim missing" case, not the
weaker "claim present but null" case.

**🟡→✅ Nit — `col_is_fk` didn't pin `ON DELETE` behavior.**
Closed by the same `confdeltype` + refused-`DELETE` pair described above — a future edit to the FK's
delete clause now fails CI instead of drifting silently.

**Assertion count re-verified by hand:** 13 original + 4 new (`confdeltype`, refused `DELETE`, no-claim
`SELECT`, no-claim `INSERT`) = 17, matching the updated `plan(17)` exactly — no silently under- or
over-declared plan.

**No pushback:** all four of the author's reply comments on the round-1 threads say "Fixed," with the
commit that did it. Nothing contested this round.

### No regressions

The fix commit (`f6cefba3`) touches exactly the four files the round-1 findings named — migration,
schema table, module README, test file. No unrelated files changed. Every file untouched by the fix
(`_shared/01_functions.sql`, `_types/enums.sql`, `02_indexes.sql`, `03_rls.sql`, top-level `README.md`)
is unchanged and still holds up on re-verification against live branch content.

<details>
<summary>Per-file re-review (round 2) — click to expand</summary>

| File | Verdict | Δ this round |
|---|---|---|
| `README.md` | safe | unchanged |
| `supabase/migrations/20260814190000_43_user_consents.sql` | safe | +4 −1 |
| `supabase/schemas/_shared/01_functions.sql` | safe | unchanged |
| `supabase/schemas/_types/enums.sql` | safe | unchanged |
| `supabase/schemas/consent/user_consents/01_table.sql` | safe | +10 −3 |
| `supabase/schemas/consent/user_consents/02_indexes.sql` | safe | unchanged |
| `supabase/schemas/consent/user_consents/03_rls.sql` | safe | unchanged (fix landed as a test, not a code change — correctly scoped, since the policy logic itself needed no fix, only proof) |
| `supabase/schemas/consent/user_consents/README.md` | safe | +18 −3 |
| `supabase/tests/database/user_consents_test.sql` | safe | +56 −4 |

**Also re-confirmed, not previously called out separately:**
- Migration and declarative schema files remain consistent with no drift (re-verified live).
- The two originally-absent constraints (no FK to `consent_versions`, no `UNIQUE` on
  `(user_id, consent_type, version)`) are unchanged and still correct — not affected by this round.
- `03_rls.sql`'s policies were already correctly written to deny an absent claim (`NULLIF` yields
  `NULL`, which never equals a `NOT NULL` column); round 1's nit was that this correct behavior was
  untested, not that the policy code was wrong.

</details>

### Suggested next steps

- [x] Nothing outstanding — ready to merge as-is.

---

**Verdict: APPROVE.** Every round-1 finding is resolved and independently verified against live branch
content at `f6cefba3` (not trusted from the diff or from reply comments alone). 0 blockers, 0 nits,
0 new findings, 0 regressions.

supabase/schemas/consent/user_consents/01_table.sql
```diff
@@ -0,0 +1,41 @@
...
+-- specifies app_version and platform as the only context captured.
+-- See README.md.
+
+CREATE TABLE IF NOT EXISTS "public"."user_consents" (
+  "id"           uuid PRIMARY KEY DEFAULT (gen_random_uuid()),
+  "user_id"      uuid NOT NULL REFERENCES "public"."users"("id"),
+  "consent_type" "public"."consent_type_enum" NOT NULL,
+  "version"      integer NOT NULL,
+  "action"       "public"."consent_action_enum" NOT NULL,
+  "recorded_at"  timestamptz NOT NULL DEFAULT (now()),
```
ayanasamuel8, line 16: **The column that orders the gate is fully client-supplied, with no guard.**

`recorded_at` is written by the client on every insert — the app's `UserConsentDto.toInsertJson()` sends `recorded_at: recordedAt.toUtc().toIso8601String()` unconditionally, so `DEFAULT now()` never actually applies. Combined with no FK on `(consent_type, version)` (deliberate, agreed) and an INSERT policy that checks only `user_id = jwt_internal_user_id()`, every field of the row that "decides the gate" — `recorded_at` (the ordering key per this table's own header comment), `version`, `action` — is authored by the client.

A client can therefore `INSERT (action='accepted', version=99, recorded_at='2099-01-01')` for its own user, and no later legitimate `withdrawn` row can ever sort above it — i.e. spoof "consent given" indefinitely, defeating a future re-consent requirement and (for `analytics`) keeping capture enabled without valid consent. That's the scenario CA-962's Security Considerations section exists to close.

There's a real reason the client must set `recorded_at` (offline-first writes via PowerSync / CA-971 can sync long after the action), so forcing server time with a `BEFORE INSERT` trigger would be wrong. A one-sided bound stops the future-dating case without breaking offline: `CHECK (recorded_at <= now() + interval '1 day')`, or a trigger that clamps a future `recorded_at` to `now()`. If it's an accepted tradeoff instead, please give it the same explicit written justification the header already gives the three deliberately-absent constraints — otherwise a reviewer keeps re-finding it.

supabase/schemas/consent/user_consents/01_table.sql
```diff
@@ -0,0 +1,41 @@
...
+  "app_version"  text,
+  "platform"     text,
+
+  -- NOT `> 0`, unlike consent_versions.version. A withdrawal records the
+  -- version it revokes, and the client writes 0 when there is nothing on file
+  -- to revoke; `> 0` would reject exactly those withdrawals at runtime.
+  CONSTRAINT "user_consents_version_non_negative" CHECK ("version" >= 0)
+);
```
ayanasamuel8, line 40: **`version >= 0` here is contradicted by the client DTO, which rejects `version <= 0` on read-back.** (Verified against `lib/libraries/consent/data/models/user_consent_dto.dart` on the CA-963 branch.)

This check deliberately allows `version = 0` (a withdrawal with nothing on file to revoke) — the comment right above and `user_consents_test.sql` both make a point of it. But `UserConsentDto.fromJson` has:

```dart
_ when version == null || version <= 0 => DatabaseConstants.versionColumn,
// Zero mirrors the backend's check (version > 0); CA-963 §2.
```

That comment applies §2's `consent_versions` rule (`version > 0`) to the §3 `user_consents` table, which is a different table with `version >= 0`. Net effect: when `ConsentRepositoryImpl.recordWithdrawal` writes `version: 0` and that row syncs back, `fromJson` throws `FormatException('Unreadable user consent row: version')`, and the repository maps that throw to a *status* — so a legitimate withdrawal silently makes the gate unreadable. The fix belongs in the app's `UserConsentDto`, not this migration, but the two sides need reconciling before the stack lands.

ayanasamuel8 (changes requested):
The append-only mechanism is implemented the right way for this codebase and I verified it independently: RLS on, exactly two policies (SELECT + INSERT, both `TO authenticated` and scoped to `user_id = jwt_internal_user_id()`), no UPDATE/DELETE policy for any role — which, with `auto_expose_new_tables`, is the real enforcement point. Naming/numbering, `OWNER TO postgres`, `timestamptz`, the `(user_id, consent_type, recorded_at DESC)` index, the bare `users(id)` FK, and the pgTAP suite (`plan(17)`, incl. the silent zero-row UPDATE/DELETE assertions) all match repo convention. Prior review items (`ON DELETE CASCADE`, the missing claim-absent test, pinning the FK's `ON DELETE`) are resolved and I'm not re-raising them; the three deliberately-absent constraints are well justified and should stay.

Two issues on the trust model of the gate-deciding columns — inline. The second is a confirmed cross-stack contradiction (verified against the app DTO on the CA-963 branch) and should be reconciled before the stack merges.

Stack-level: CA-963/CA-964 both instruct that this stack not merge until CA-962's Approver sign-off lands and the two open design-doc comments (one questioning analytics as an independently-withdrawable consent type — now baked into `consent_type_enum`) are resolved. Worth confirming before #48 → `master`.

PR #50, opened 2026-08-18
CA-845: chore(deps): bump Supabase CLI pin to 2.106.0 to match CI

Bazabizi:
## Context

Second half of [CA-845](https://ripplearc.youtrack.cloud/issue/CA-845) (the app-side dead-datasource deletion already landed via CA-838 R1, app PR #466).

`devDependencies.supabase` pinned `^2.22.12`, which fails to parse the current `supabase/config.toml` (`auto_expose_new_tables`, introduced for CA-729). Local dev has been bypassing the pin with `npx supabase@latest`.

## Change

- Pin `supabase` to exactly `2.106.0` — the same version CI installs via `supabase/setup-cli` in `database_tests.yml` — so local dev and CI run the identical CLI. An exact pin (vs caret) is deliberate: `^2.106.0` would silently float past the CI version again.
- `package-lock.json` / `yarn.lock` re-synced by npm.

## Verification

- `npx supabase --version` → `2.106.0`
- `npx supabase status` parses the config cleanly (the old pin hard-failed here)

## Notes

- 2.106.0 warns that `api.auto_expose_new_tables` is deprecated and will be removed on 2026-10-30. Removing it changes Data API exposure defaults, so that belongs with the CA-949 hardening work, not this chore.
- No README changes: all documented `npx supabase ...` commands are unchanged and now resolve to the pinned version.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,340 +5,151 @@
   "packages": {
     "": {
       "devDependencies": {
-        "supabase": "^2.22.12"
+        "supabase": "2.106.0"
       }
     },
-    "node_modules/@isaacs/fs-minipass": {
-      "version": "4.0.1",
-      "resolved": "https://registry.npmjs.org/@isaacs/fs-minipass/-/fs-minipass-4.0.1.tgz",
-      "integrity": "sha512-wgm9Ehl2jpeqP3zw/7mo3kRHFp5MEDhqAdwy1fTGkHAwnkGOVsgpvQhL8B5n1qlb01jV3n/bI0ZfZp5lWA1k4w==",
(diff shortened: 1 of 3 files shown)


Bazabizi:
## 🔍 Self-review (applied at PR creation)

✅ No issues found. The diff is the single pinned version line plus npm-synced lockfiles; the exact-pin-vs-caret rationale, CI alignment, the `auto_expose_new_tables` deprecation note (CA-949), and verification commands are documented in the PR body. `yarn.lock` changes are npm ≥7's automatic lockfile sync, not a package-manager switch.

PR #51, opened 2026-08-19
CA-749: seed the auth credential for the sample user

ayanasamuel8:
### 1. PR SUMMARY

The sample data seeded a `public.users` row for `seeder@example.com` whose
`credential_id` referenced an `auth.users` id that was never created, and no
foreign key caught the dangling reference. The account could be read from the
database but could never sign in, so end-to-end sign-in journeys had no usable
user. This seeds the matching GoTrue credential and its email identity, and adds
a pgTAP test that fails if the link breaks again.

**Type:** Feature ·
**Size:** S (69 seeder lines, 54 of them statements; 41 test lines excluded) ·
**Rules applied:** Digestible PR

> **Sizing note:** the rule's detection pattern counts only `lib/**/*.dart`, which is zero here — that would report XS and understate the change. The sizes below are the branch's actual reviewable lines (tests and generated files excluded).

### What changed
- `supabase/seeders/sample_data/100_auth_users.sql` — the `auth.users` credential and its `auth.identities` row, numbered to load before `103_users.sql`.
- `supabase/tests/database/seeded_auth_user_test.sql` — 5 assertions covering existence, the `credential_id` link, email confirmation, password verification and the email identity.

### 2. REVIEW SUMMARY TABLE

> Found 2 issue(s): 0 🔴 critical · 0 🟠 major · 0 🟡 minor · 2 🔵 suggestion — across 1 file(s).

| # | Issue | Severity | Location | Description | Suggested Fix | Status | Notes |
|---|-------|----------|----------|-------------|---------------|--------|-------|
| 1 | Signable account reaches every local database | 🔵 Suggestion | `supabase/seeders/sample_data/100_auth_users.sql:1` | Placing this in shared `sample_data/` means local development and `database_tests.yml` also get an account with a published password, not just the E2E stack. | Accepted by design — one seeded fixture shared by all three consumers is the point, and the credential is only ever valid against a local Docker stack. The password comment states that constraint explicitly. | ❌ Disagree | Approved placement; splitting it into an E2E-only seed would let the two drift. |
| 2 | Verification depended on unstated schema assumptions | 🔵 Suggestion | `supabase/seeders/sample_data/100_auth_users.sql:50` | `auth.identities` gained `provider_id` in a GoTrue schema change, and `crypt`/`gen_salt` live in the `extensions` schema rather than `public` — either assumption being wrong would fail the seed at load time. | Verified on an isolated stack: the seeders loaded in order and the full suite passed (19 files, 197 tests). | ✅ Addressed | |

**Status (author fills):** ✅ Addressed · 📋 Future Story (add YouTrack ID) ·
❌ Disagree (justify) · 🔄 In Progress

## Test plan

Verified on an **isolated** Supabase stack (`project_id = "ca749probe"`, ports 556xx) built from a copy of this repo's config — never against a developer's real local stack.

- Seeders load in the intended order: `100_auth_users.sql` runs before `103_users.sql`.
- `auth.users` row exists, email is confirmed, and the password verifies via `extensions.crypt`.
- `auth.identities` email identity present; `public.users.credential_id` equals `auth.users.id`.
- Full pgTAP suite: `Files=19, Tests=197 ... Result: PASS`, including the new `seeded_auth_user_test.sql`, with no existing test regressed.
- Seeds are idempotent across `supabase db reset` — the password still verifies afterwards.

### DIFF
--- a/supabase/seeders/sample_data/100_auth_users.sql
+++ b/supabase/seeders/sample_data/100_auth_users.sql
@@ -1,0 +1,69 @@
+-- Auth Users Seeder
+-- Creates the GoTrue account backing the seeded public.users row in
+-- 103_users.sql. public.users.credential_id points at auth.users.id, but no
+-- foreign key enforces it, so without this file the seeded user can be read
+-- from the database yet can never sign in. E2E sign-in flows depend on it.
+--
+-- Runs before 103_users.sql so the credential exists before it is referenced.
+
+-- The seeded password is a fixed, well-known value. It is only ever valid
+-- against a local Docker Supabase stack started from this repository, which is
+-- never exposed outside the developer machine or the CI runner. Never reuse it
+-- for a hosted environment.
+INSERT INTO auth.users (
+  "instance_id",
+  "id",
+  "aud",
+  "role",
+  "email",
+  "encrypted_password",
+  "email_confirmed_at",
+  "raw_app_meta_data",
+  "raw_user_meta_data",
+  "created_at",
+  "updated_at",
+  "confirmation_token",
+  "recovery_token",
+  "email_change_token_new",
+  "email_change"
+) VALUES
+  (
+    '00000000-0000-0000-0000-000000000000',
+    '850e8400-e29b-41d4-a716-446655440000',
+    'authenticated',
+    'authenticated',
+    'seeder@example.com',
+    extensions.crypt('e2e-local-only-password', extensions.gen_salt('bf')),
+    now(),
+    '{"provider": "email", "providers": ["email"]}',
+    '{}',
+    now(),
+    now(),
+    '',
+    '',
+    '',
+    ''
+  )
+ON CONFLICT ("id") DO NOTHING;
+
+-- GoTrue resolves the email/password grant through auth.identities, so the
+-- account is not usable without a matching email identity.
+INSERT INTO auth.identities (
+  "provider_id",
+  "user_id",
+  "identity_data",
+  "provider",
+  "last_sign_in_at",
+  "created_at",
+  "updated_at"
+) VALUES
+  (
+    '850e8400-e29b-41d4-a716-446655440000',
+    '850e8400-e29b-41d4-a716-446655440000',
+    '{"sub": "850e8400-e29b-41d4-a716-446655440000", "email": "seeder@example.com", "email_verified": true, "phone_verified": false}',
+    'email',
+    now(),
+    now(),
+    now()
+  )
+ON CONFLICT ("provider_id", "provider") DO NOTHING;
--- a/supabase/tests/database/seeded_auth_user_test.sql
+++ b/supabase/tests/database/seeded_auth_user_test.sql
@@ -1,0 +1,41 @@
+begin;
+select plan(5);
+
+-- The seeded public.users row is unusable for sign-in unless a matching GoTrue
+-- credential exists, which is what these assertions guard.
+
+SELECT ok(
+  EXISTS (SELECT 1 FROM auth.users WHERE "email" = 'seeder@example.com'),
+  'seeded auth user should exist'
+);
+
+SELECT is(
+  (SELECT "credential_id" FROM public.users WHERE "email" = 'seeder@example.com'),
+  (SELECT "id" FROM auth.users WHERE "email" = 'seeder@example.com'),
+  'public.users.credential_id should match the seeded auth user id'
+);
+
+SELECT ok(
+  (SELECT "email_confirmed_at" IS NOT NULL FROM auth.users WHERE "email" = 'seeder@example.com'),
+  'seeded auth user should have a confirmed email'
+);
+
+SELECT ok(
+  (
+    SELECT "encrypted_password" = extensions.crypt('e2e-local-only-password', "encrypted_password")
+    FROM auth.users WHERE "email" = 'seeder@example.com'
+  ),
+  'seeded auth user password should verify against the documented seed password'
+);
+
+SELECT ok(
+  EXISTS (
+    SELECT 1 FROM auth.identities i
+    JOIN auth.users u ON u."id" = i."user_id"
+    WHERE u."email" = 'seeder@example.com' AND i."provider" = 'email'
+  ),
+  'seeded auth user should have an email identity'
+);
+
+select * from finish();
+rollback;


ripplearcgit (changes requested):
## Review — CA-749: seed the auth credential for the sample user

**Verdict: Request changes** — 2 blocking, 1 major. To be clear about what is *not* at issue: the security question this PR raises resolves cleanly, and the SQL itself is correct. Both blockers are about what needs to land alongside it.

### Architectural overview

```mermaid
sequenceDiagram
    participant Dev as Developer or CI
    participant CLI as Supabase CLI
    participant Local as Local Docker stack
    participant Hosted as Hosted Supabase project

    rect rgba(139,92,246,0.15)
    Note over Dev,Hosted: Analogy: the spare key goes under the doormat of the practice house only
    end

    Dev->>CLI: supabase db reset / db start
    CLI->>Local: apply supabase/migrations
    CLI->>Local: run seeders/**/*.sql (100 then 103)
    Local-->>Dev: seeder@example.com can now sign in
    rect rgba(139,92,246,0.15)
    Note over Local: the key sits under this doormat
    end

    Dev->>CLI: supabase db push
    CLI->>Hosted: apply supabase/migrations ONLY
    Note over CLI,Hosted: seeders are not part of db push
    rect rgba(139,92,246,0.15)
    Note over Hosted: the real plans have no doormat step
    end

    Dev->>CLI: supabase db reset --linked (destructive, manual)
    CLI->>Hosted: attempt seeders as role postgres
    Hosted-->>CLI: INSERT on auth.users refused (SELECT-only grant)
    rect rgba(139,92,246,0.15)
    Note over Hosted: the real door will not take the spare key
    end
```

The seeded credential is reachable only from a local stack. Seeders run on `supabase start` and `supabase db reset`; the repo has no deploy workflow; and `supabase db push` — the command that talks to a hosted project — applies `supabase/migrations/` only and never reads `supabase/seeders/`. The one human-reachable escape hatch, `db reset --linked`, would hit a second independent wall: GoTrue's own `20240612123726_enable_rls_update_grants` grants the `postgres` role only `SELECT` on `auth.users` and `auth.identities`.

I checked the risky schema assumptions against GoTrue's migration files rather than taking the PR body's word for it — `confirmed_at` correctly omitted (generated), `identities.id` correctly omitted (has a default), `identities.email` correctly omitted (generated), `ON CONFLICT (provider_id, provider)` names a real constraint, and `extensions.crypt` correctly qualified. Every one of them is right. The seeder ordering contract holds too: globs expand alphabetically, so `100_` loads before `103_`.

---

## 🔴 Blocking

### 1. File the follow-up ticket for the missing foreign key

The root cause named in this PR's own summary is still unfixed: `public.users.credential_id` has no foreign key to `auth.users.id`. The PR body says it plainly — *"no foreign key caught the dangling reference."* This PR fixes the instance and adds a regression test, but the class of bug stays creatable by anyone who writes a `public.users` row by hand.

```mermaid
flowchart LR
    A["auth.users.id = 850e8400 ... 440000"]
    B["public.users.credential_id = 850e8400 ... 440000"]
    A -. "no FOREIGN KEY constraint" .-> B
    A -- "asserted only by seeded_auth_user_test" --> B
    style A fill:#e8f0fe,stroke:#4285f4
    style B fill:#e6f4ea,stroke:#34a853
```

**The FK itself belongs in a follow-up, not in this diff.** Adding `REFERENCES auth.users(id)` would immediately break eight existing pgTAP files — `cascade_delete_test`, `cost_estimates_test`, `cost_items_test`, `cost_estimate_logs_test`, `cost_estimate_activity_logging_test`, `cost_estimates_update_guard_test`, `critical_tables_insert_and_columns_test` and the search-function tests — all of which insert `public.users` rows with invented `credential_id` values like `22222222-2222-2222-2222-222222222222`. Making the link real means giving every one of those a matching credential first.

**What blocks is the tracking, not the fix.** Please open a YouTrack ticket before this merges, covering (a) the FK itself, (b) the cascade choice — `ON DELETE CASCADE` is Supabase's own convention here, and (c) the test-fixture migration that has to precede it. The bug this PR's summary names shouldn't merge untracked.

### 2. Close the GoTrue version gap

This seeder hard-codes the shape of a schema the repo does not own. The `auth` schema is migrated by the GoTrue container, not by `supabase/migrations/`, and GoTrue is still actively changing it — its most recent migration is dated 2026-06, and it has renamed a column in exactly this table before (`identities.id` → `identities.provider_id`, migration `20231117164230_add_id_pkey_identities`). That is precisely the failure mode your own PR body flags as *"Verification depended on unstated schema assumptions."*

The pinning is inconsistent, which widens the window: `package.json` pins the CLI at `supabase@^2.22.12` for local `npx` use, while `.github/workflows/database_tests.yml` installs `supabase/setup-cli@v1` with `version: latest`. Local and CI can therefore boot different GoTrue images. When a future release adds a `NOT NULL` column or renames one, the symptom is `supabase db reset` failing for every developer at once, with a seed-time error that reads as unrelated to any code change.

**Either option satisfies this — your call:**
- Pin the CI workflow to the same CLI version as `package.json`, so local and CI fail together rather than mysteriously apart. One-line change, gets most of the value.
- Or create the account through GoTrue's Admin API in a setup script instead of raw SQL, making GoTrue responsible for its own schema. More work, arguably overkill for a fixture.

Leaving both undone is what blocks.

---

## 🟠 Major — README still documents the procedure this PR replaces

`README.md` §"Linking the Test User to Supabase Auth" (lines 414–424, plus the ToC entry at line 23) still walks developers through creating a second auth user in Studio and pasting its UID over `credential_id`. After this merges, anyone following it hits one of two outcomes, neither pointing back at the README as the cause:

```mermaid
flowchart TD
    A["Dev follows README: Linking the Test User to Supabase Auth"] --> B["Studio: Add User with seeder@example.com"]
    B --> C{"auth.users now already holds that email"}
    C -->|GoTrue rejects the signup| D["Confusing error: user already registered"]
    C -->|Dev pastes a fresh UID anyway| E["public.users.credential_id overwritten"]
    E --> F["seeded_auth_user_test assertion 2 fails"]
    E --> G["RLS returns empty rows for the seeded account"]
    style D fill:#fef9c3,stroke:#ca8a04
    style F fill:#fce8e6,stroke:#d93025
    style G fill:#fce8e6,stroke:#d93025
```

Two more spots are now incomplete: the Sample Data seeder table (line 245) is missing `100_auth_users`, and the Test Coverage table (line 263) is missing `seeded_auth_user_test`.

This is a README-only edit, not a code change — but it should land with this PR rather than after it, because the two documents actively contradict each other the moment this merges.

supabase/seeders/sample_data/100_auth_users.sql
```diff
@@ -0,0 +1,69 @@
...
+    '{"provider": "email", "providers": ["email"]}',
+    '{}',
+    now(),
+    now(),
+    '',
+    '',
+    '',
+    ''
+  )
+ON CONFLICT ("id") DO NOTHING;
```
ripplearcgit: 🟡 **`ON CONFLICT ("id")` misses the unique-email index**

The conflict target here only covers the primary key. `auth.users` also carries `users_email_partial_key`, a partial unique index on `email`. If a row with `seeder@example.com` already exists under a *different* id — which is exactly what happens to anyone who previously followed the README's manual Studio procedure — this insert aborts the whole seed run rather than skipping.

A targetless form covers both:

```sql
ON CONFLICT DO NOTHING;
```

One word, and re-seeding stays idempotent for developers with an existing hand-made account.

ripplearcgit (commented):
## Nits and notes

Non-blocking. The two inline comments above are the actionable ones; everything below is context that has no other route to you.

### 🟡 Nits
- `100_auth_users.sql:47` — `ON CONFLICT ("id")` misses `users_email_partial_key`; a targetless `ON CONFLICT DO NOTHING` covers both. *(inline)*
- `seeded_auth_user_test.sql:31` — the identity assertion never checks `provider_id`, so a typo'd UUID passes all five assertions. *(inline)*

### 🟢 Notes
- `100_auth_users.sql:7` — the comment says the file *"Runs before 103_users.sql so the credential exists before it is referenced."* Nothing enforces that reference yet (that's blocker 1); worth softening to say the ordering is a convention the seeder loader guarantees, not a constraint the database does.
- `100_auth_users.sql:49–50` — the comment says GoTrue *"resolves the email/password grant through auth.identities."* The identity row is required for a complete account, but it isn't what resolves the password grant — `encrypted_password` on `auth.users` is. Worth rewording so the next reader doesn't infer a dependency that isn't there.
- The repo is **public**, so `e2e-local-only-password` is world-readable. That's acceptable given the credential is only ever valid against a local Docker stack — but it raises the stakes if Supabase preview branches are ever enabled, since those auto-run seeders on internet-reachable projects. Not configured today; worth knowing before it is.

<details>
<summary><b>Risk map — what was scrutinised</b></summary>

| File | Change | Verdict |
|---|---|---|
| `supabase/seeders/sample_data/100_auth_users.sql` | +69 | 🟠 Amber — not because it is wrong (it verifies clean against the real GoTrue schema) but because it hard-codes a table this repo does not own, and plants a working credential three consumers now load |
| `supabase/tests/database/seeded_auth_user_test.sql` | +41 | 🟡 Sound; one assertion gap (`provider_id`) |

</details>

<details>
<summary><b>On "Suggestion 1 · Disagree" — your call is right</b></summary>

Your PR body raises, then rejects, the objection that putting this credential in shared `sample_data/` gives every local database and every CI run a signable account, not just the E2E stack. I agree, and sibling #52 doesn't change that calculus.

\#52 renames the local `project_id` so the E2E stack's Docker containers and volumes stop colliding with a developer's own. That is container namespacing — it says nothing about which SQL files get loaded. Both stacks still load the same `seeders/**/*.sql` set. Splitting this into an E2E-only seed would recreate exactly the drift CA-749 exists to fix, and give the two copies room to disagree again.

Worth noting your own test plan used the right recipe for isolation — a distinct `project_id` **and** distinct ports (`ca749probe`, 556xx), in a copy of the repo.

</details>

### Suggested next steps

- [ ] Open the FK follow-up ticket — `credential_id` → `auth.users(id)`, the `ON DELETE CASCADE` choice, and the eight pgTAP fixtures that must migrate first 🔴
- [ ] Pin `database_tests.yml` to `package.json`'s CLI version, **or** move account creation to GoTrue's Admin API 🔴
- [ ] `README.md` 414–424 + ToC line 23 — delete the manual "Linking the Test User to Supabase Auth" procedure 🟠
- [ ] `README.md` line 245 — add `100_auth_users` to the Sample Data seeder table 🟠
- [ ] `README.md` line 263 — add `seeded_auth_user_test` to the Test Coverage table 🟠
- [ ] `100_auth_users.sql:47` — targetless `ON CONFLICT DO NOTHING` 🟡
- [ ] `seeded_auth_user_test.sql:31` — assert `provider_id`; bump the plan to 6 🟡
- [ ] `100_auth_users.sql:7` and `49–50` — soften the ordering comment, reword the identity comment 🟢

supabase/tests/database/seeded_auth_user_test.sql
```diff
@@ -0,0 +1,41 @@
...
+
+SELECT ok(
+  (
+    SELECT "encrypted_password" = extensions.crypt('e2e-local-only-password', "encrypted_password")
+    FROM auth.users WHERE "email" = 'seeder@example.com'
+  ),
+  'seeded auth user password should verify against the documented seed password'
+);
+
+SELECT ok(
```
ripplearcgit, line 31: 🟡 **The identity assertion can pass with a wrong `provider_id`**

This checks that an email identity *exists* and joins to the right user, but never asserts `i."provider_id"` equals the auth user's id. `provider_id` is a plain `text` column with no foreign key, so a typo'd UUID in the seeder would still satisfy all five assertions — the test would go green on a fixture that cannot actually sign in, which is the precise failure this file exists to catch.

Suggested addition (and bump `plan(5)` → `plan(6)`):

```sql
SELECT is(
  (SELECT i."provider_id" FROM auth.identities i
     JOIN auth.users u ON u."id" = i."user_id"
    WHERE u."email" = 'seeder@example.com' AND i."provider" = 'email'),
  '850e8400-e29b-41d4-a716-446655440000',
  'seeded identity provider_id should match the auth user id'
);
```

supabase/seeders/sample_data/100_auth_users.sql
ayanasamuel8: Fixed in b6b3b91 — switched to a targetless `ON CONFLICT DO NOTHING`. Added a comment explaining why (covers `users_email_partial_key`, not just the PK, so a developer who'd already hand-created the account via the old manual Studio procedure gets a no-op instead of an aborted seed run). Resolving.

supabase/tests/database/seeded_auth_user_test.sql
ayanasamuel8: Fixed in b6b3b91 — added the `provider_id` assertion exactly as suggested and bumped `plan(5)` to `plan(6)`. Ran the full suite twice (reset -> test -> reset -> test) to confirm it's green and idempotent: `Files=19, Tests=198 ... Result: PASS` both times, on the now-pinned CLI 2.106.0. Resolving.

ayanasamuel8:
Replying to the non-inline findings from the two review comments (both Blockers, the Major, and the two 🟢 Notes), since GitHub won't let me thread a reply directly under them. Pushed in b6b3b91.

**🔴 Blocker 1 — file the FK follow-up ticket.** Filed [CA-995](https://ripplearc.youtrack.cloud/issue/CA-995) (subtask of CA-749, Task, P2, 2pts): adds `public.users.credential_id -> auth.users.id`, decides `ON DELETE CASCADE` vs `RESTRICT`, and lists all 8 pgTAP fixtures that need migrating off invented `credential_id` values first, exactly as you scoped it. Referenced it in the seeder's ordering comment too so the next reader has a pointer to why the ordering isn't (yet) enforced by the schema.

**🔴 Blocker 2 — close the GoTrue version gap.** Took the one-line option: pinned `package.json`'s `supabase` devDependency to the exact version CI uses (`2.106.0`, was a floating `^2.22.12`), and regenerated `package-lock.json`/`yarn.lock`. Also caught and fixed the stray `.github/database_tests.yml` (outside `workflows/`, so it's dead — GitHub only runs workflows from `.github/workflows/` — but still pinned to `latest`, which would've been misleading if anyone copies from it). `npx supabase --version` now reports `2.106.0` locally, matching CI exactly. Verified `supabase db reset` + `supabase test db` on that pinned version, twice in a row (idempotency check) — `Files=19, Tests=198 ... Result: PASS` both times.

**🟠 Major — README still documents the replaced procedure.** Replaced the "Linking the Test User to Supabase Auth" manual Studio walkthrough with a short "Signing In as the Seeded Test User" section stating the credentials are ready out of the box now. Updated the ToC entry, added `100_auth_users` to the Sample Data seeder table, and added `seeded_auth_user_test` to the Test Coverage table.

**🟢 Note — ordering comment overstated what's enforced.** Reworded: now says the ordering is the seeder loader's alphabetical convention, not a DB constraint, and points at CA-995 for the actual enforcement gap.

**🟢 Note — identity comment implied the wrong thing resolves the password grant.** Reworded: now says `encrypted_password` on `auth.users` is what verifies the password, and the identity row is required for a *complete* account rather than for password resolution itself.

**🟢 Note — world-readable password / preview branches.** No action taken, per your own framing — it's a "worth knowing," not a blocking item, and preview branches aren't configured today. Noting it here so it's visible if that ever changes.

**On "Suggestion 1 · Disagree" / the sibling-PR risk-calculus note:** no disagreement from me either — leaving the shared `sample_data/` placement as-is, agreed splitting it E2E-only would just reintroduce the drift CA-749 fixes.

README.md
```diff
@@ -411,16 +413,14 @@ When testing signup or password reset, emails are intercepted by **Inbucket** (n
...
 
-1. **Create the auth user** — In Studio (`http://localhost:54323`) → Authentication → Users → Add User → enter email `seeder@example.com` and a password.
-2. **Copy the User UID** from the auth users list.
-3. **Update the users table** — In Table Editor → `users` table → find `seeder@example.com` → paste the UID into the `credential_id` field.
-4. **Verify** — Log in through the app. RLS policies should now work correctly.
+- **Email:** `seeder@example.com`
+- **Password:** `e2e-local-only-password`
 
-> **Why?** The `credential_id` column links a `users` row to its Supabase Auth identity (`auth.uid()`). Without this match, RLS policies will deny access.
+This password is fixed and committed to this public repo — it is only ever valid against a local Docker Supabase stack started from this repository, never a hosted environment. `public.users.credential_id` already matches this account's `auth.users.id`, so RLS policies work correctly without any extra linking step.
```
ripplearcgit, line 423: 🟠 **Major 1 — this hunk collides head-on with #52, and one resolution silently undoes R1's Major**

Both PRs branch off the same `master` commit and both edit this exact region. **#51 deletes** the manual Studio procedure at lines 413–424 and replaces it. **#52 keeps** that procedure and edits inside it, rewriting `http://localhost:54323` to `55323` as part of its port renumbering. That is a modify/delete overlap on the same hunk — git will conflict whichever order they merge in.

The risk is not the conflict itself, it is how it gets resolved. If someone resolving it takes **#52's** side of the hunk — which looks like the safe choice, because #52's version is the one with the "newer" port numbers — the four-step manual procedure comes straight back, and the exact contradiction round 1 blocked on is live again with nothing left to catch it. The reverse order is milder but not free: taking #51's side discards #52's port edit inside that section with no signal, though since the section itself is being deleted, nothing is actually lost.

```mermaid
flowchart TD
    B["README.md on master — lines 413-424 hold the manual Studio procedure"]
    B --> P51["#51 DELETES the whole section and replaces it with the seeded credentials"]
    B --> P52["#52 KEEPS the section, only rewrites 54323 to 55323 inside it"]
    P51 --> M{"Both target master — same lines"}
    P52 --> M
    M -->|"#51 merges first, then #52"| C1["git conflict on that hunk"]
    C1 --> R1["Resolve by keeping #51: correct"]
    C1 --> R2["Resolve by keeping #52: the deleted manual procedure comes back"]
    M -->|"#52 merges first, then #51"| C2["git conflict on that hunk"]
    C2 --> R3["#51 wins the hunk: correct, but #52 port renumbering inside it is lost with no signal"]
    style P51 fill:#e6f4ea,stroke:#34a853
    style P52 fill:#e8f0fe,stroke:#4285f4
    style R2 fill:#fce8e6,stroke:#d93025
    style R3 fill:#fef9c3,stroke:#ca8a04
```

**Ask:** state the merge order on both PRs — merge **#51 first**, then rebase #52 and drop its `54323 → 55323` edit inside "Linking the Test User to Supabase Auth" entirely, since the section no longer exists. Cheap to say now, expensive to notice later: nothing in CI tests the README, so a bad resolution ships silently.

.github/database_tests.yml
```diff
@@ -9,6 +9,6 @@ jobs:
       - uses: actions/checkout@v3
       - uses: supabase/setup-cli@v1
         with:
-          version: latest
+          version: 2.106.0
```
ripplearcgit: 🟡 **Minor 3 — this file is dead config; pinning it institutionalises a second copy**

GitHub only executes workflow files under `.github/workflows/`. This one sits directly in `.github/`, so it has never run and never will — the PR body calls it "the stray/unused legacy workflow copy," which is accurate. It is a nine-line skeleton missing the skip-CI guard, the signing-key step and the path filters that the real workflow has, so it is also stale, not just inert.

Pinning it makes it look maintained. The next person to change the CLI version now has two files to remember, one of which does nothing, and a future reader has no way to tell which is authoritative without knowing GitHub's directory rule. **Suggest deleting it** in this PR instead — it is a one-line change in the opposite direction and removes the ambiguity permanently.

```
git rm .github/database_tests.yml
```

Not blocking. If there is a reason it is being kept that I cannot see from the repo, a one-line comment at the top saying so would be enough.

yarn.lock
```diff
@@ -2,165 +2,21 @@
...
+  integrity sha512-jV0PJi5LqXUMZ4kriRLvST3DX1YDhrxOgnFJLOn6oUw3KQEFJPXhA60yLQ9aWG1E1uZCp4vUJwSuSGq3HPfaSQ==
+  optionalDependencies:
+    "@supabase/cli-darwin-arm64" "2.106.0"
+    "@supabase/cli-darwin-x64" "2.106.0"
+    "@supabase/cli-linux-arm64" "2.106.0"
+    "@supabase/cli-linux-arm64-musl" "2.106.0"
+    "@supabase/cli-linux-x64" "2.106.0"
+    "@supabase/cli-linux-x64-musl" "2.106.0"
+    "@supabase/cli-windows-arm64" "2.106.0"
+    "@supabase/cli-windows-x64" "2.106.0"
```
ripplearcgit: 🟡 **Minor 2 — the regenerated `yarn.lock` only resolves the Linux x64 binary**

Same investigation as `package-lock.json`, different answer. The `supabase@2.106.0` stanza correctly *declares* all eight platform packages as `optionalDependencies`, but the file contains a resolved entry for exactly one of them: `@supabase/cli-linux-x64`. `package-lock.json`, generated from the same `package.json`, carries all eight. The two lockfiles now disagree about the same dependency set, which is a strong hint this one was regenerated inside a Linux container while the npm one was not.

What it costs, concretely: a developer on Apple Silicon running `yarn install --frozen-lockfile` has no `@supabase/cli-darwin-arm64` entry to install from and gets a lockfile-out-of-date failure. Running plain `yarn install` instead succeeds but rewrites `yarn.lock` on their machine, so the file churns back and forth in unrelated PRs depending on who last touched it.

```mermaid
flowchart TD
    P["supabase@2.106.0 declares 8 platform binaries as optionalDependencies"]
    P --> PL["package-lock.json — all 8 present, each with its own os and cpu fields"]
    P --> YL["yarn.lock — only @supabase/cli-linux-x64 resolved"]
    YL --> M1["macOS arm64 dev runs yarn install --frozen-lockfile"]
    M1 --> M2["no darwin-arm64 entry: lockfile-out-of-date failure"]
    YL --> M3["macOS arm64 dev runs plain yarn install"]
    M3 --> M4["yarn.lock silently rewritten: churn in every PR from that machine"]
    style PL fill:#e6f4ea,stroke:#34a853
    style YL fill:#fef9c3,stroke:#ca8a04
    style M2 fill:#fce8e6,stroke:#d93025
    style M4 fill:#fef9c3,stroke:#ca8a04
```

**The simpler fix is probably to delete `yarn.lock` entirely.** Nothing in this repo uses yarn — the README says `npm install`, CI runs no package manager at all, and `package-lock.json` is committed and complete. Two lockfiles for one dependency is a standing invitation for exactly this kind of disagreement. If yarn is deliberately supported for someone, regenerating the file on a machine that resolves all eight platforms is the alternative.

Not blocking: today CI never installs from either lockfile, so nothing breaks in the pipeline. It bites individual developers, quietly.

.github/database_tests.yml
ayanasamuel8: Fixed in 64dad6e — deleted `.github/database_tests.yml` as suggested rather than pinning it. Resolving.

yarn.lock
ayanasamuel8: Fixed in 64dad6e — deleted `yarn.lock`. Nothing in the repo or CI uses yarn and `package-lock.json` already covers all eight platforms, so the simpler fix was the right call. Resolving.

README.md
ayanasamuel8: Agreed — **#51 merges first.** Stating the order here and on #52: once #51 lands, #52 should rebase and drop its `54323 → 55323` edit inside "Linking the Test User to Supabase Auth" entirely, since that section no longer exists after this PR. Cross-posted on #52 as well so it's visible from both sides.

PR #52, opened 2026-08-20
CA-991: isolate the E2E project_id from ordinary local dev

ayanasamuel8:
## Summary
- Renames the local Supabase `project_id` from `construculator-backend` to `construculator-backend-e2e` in `supabase/config.toml`, so the containers, volumes and Docker network it creates are distinct from whatever project_id a developer's own local Supabase work defaults to.
- Updates the hardcoded network/host names derived from that project_id in `powersync/compose.yaml` (the `external: true` network reference) and `powersync/.env.example` (`PS_POSTGRES_URI`, `PS_JWKS_URI`).
- Updates the `docker exec supabase_db_...` examples in `powersync/README.md` and the two `global_search` schema READMEs so they keep working against the renamed container.

## Context
CA-749 (in `construculator-app`) stood up scripts that start/stop/reset this stack for E2E tests, assuming it was already isolated. Independent review found it wasn't: `project_id = "construculator-backend"` collided with whatever a developer's own local Supabase project defaults to from the same directory name, so `reset_env.sh` / `stop_env.sh --purge` could destroy real local dev data. Companion PR in `construculator-app`: https://github.com/ripplearc/construculator-app/pull/576 (points the app-side scripts at the renamed `construculator-backend-e2e` project).

## Test plan
- [x] `docker compose config` parity check (parsed `powersync/compose.yaml` via PyYAML) — no syntax errors, single `supabase_network_construculator-backend-e2e` network defined and referenced consistently.
- [ ] Manual: `supabase start` + `docker compose -f powersync/compose.yaml up -d` and confirm containers/network are named `..._construculator-backend-e2e`.

https://claude.ai/code/session_015gw6KVFcjiZ39MHNwcFAco

### DIFF
--- a/powersync/.env.example
+++ b/powersync/.env.example
@@ -5,5 +5,5 @@
 PS_API_TOKEN=your-secure-api-token-here
 
 PS_PORT=8081
-PS_POSTGRES_URI=postgresql://postgres:postgres@supabase_db_construculator-backend:5432/postgres
-PS_JWKS_URI=http://supabase_kong_construculator-backend:8000/auth/v1/.well-known/jwks.json
+PS_POSTGRES_URI=postgresql://postgres:postgres@supabase_db_construculator-backend-e2e:5432/postgres
+PS_JWKS_URI=http://supabase_kong_construculator-backend-e2e:8000/auth/v1/.well-known/jwks.json
--- a/powersync/README.md
+++ b/powersync/README.md
@@ -34,13 +34,13 @@ docker compose up -d
 ### 4. Verify Connection
 ```bash
 # Check replication is active
-docker exec supabase_db_construculator-backend psql -U postgres -c \
+docker exec supabase_db_construculator-backend-e2e psql -U postgres -c \
   "SELECT slot_name, active FROM pg_replication_slots WHERE slot_name LIKE 'powersync%';"
 
 # Should show: active = t
 
 # Test data sync
-docker exec supabase_db_construculator-backend psql -U postgres -d postgres -c \
+docker exec supabase_db_construculator-backend-e2e psql -U postgres -d postgres -c \
   "INSERT INTO professional_roles (name) VALUES ('Test Role') RETURNING *;"
 
 # Check PowerSync processed it
@@ -56,8 +56,8 @@ All variables are required in `.env`:
 |----------|-------------|---------|
 | `PS_API_TOKEN`| API authentication token | Generate: `openssl rand -hex 32` |
 | `PS_PORT` | PowerSync service port | `8081` |
-| `PS_POSTGRES_URI` | PostgreSQL connection string | `postgresql://postgres:postgres@supabase_db_construculator-backend:5432/postgres` |
-| `PS_JWKS_URI`  | Supabase Auth JWKS endpoint | `http://supabase_kong_construculator-backend:8000/auth/v1/.well-known/jwks.json` |
+| `PS_POSTGRES_URI` | PostgreSQL connection string | `postgresql://postgres:postgres@supabase_db_construculator-backend-e2e:5432/postgres` |
+| `PS_JWKS_URI`  | Supabase Auth JWKS endpoint | `http://supabase_kong_construculator-backend-e2e:8000/auth/v1/.well-known/jwks.json` |
 
 ### Local Development
 
@@ -177,11 +177,11 @@ docker compose logs powersync
 **No data syncing:**
 ```bash
 # 1. Check replication slot is active
-docker exec supabase_db_construculator-backend psql -U postgres -c \
+docker exec supabase_db_construculator-backend-e2e psql -U postgres -c \
   "SELECT slot_name, active FROM pg_replication_slots WHERE slot_name LIKE 'powersync%';"
 
 # 2. Verify table is in publication
-docker exec supabase_db_construculator-backend psql -U postgres -c \
+docker exec supabase_db_construculator-backend-e2e psql -U postgres -c \
   "SELECT tablename FROM pg_publication_tables WHERE pubname = 'powersync';"
 
 # 3. Check table is in sync rules (sync-config.yaml)
--- a/powersync/compose.yaml
+++ b/powersync/compose.yaml
@@ -8,7 +8,7 @@ services:
     volumes:
       - mongo_storage:/data/db
     networks:
-      - supabase_network_construculator-backend
+      - supabase_network_construculator-backend-e2e
 
   # Initialize MongoDB replica set
   mongo-rs-init:
@@ -24,7 +24,7 @@ services:
         {} rs.initiate({_id: \"rs0\", version: 1, members: [{ _id: 0, host :
         \"mongo:27017\" }]})'"
     networks:
-      - supabase_network_construculator-backend
+      - supabase_network_construculator-backend-e2e
 
   powersync:
     image: journeyapps/powersync-service:1.20.5
@@ -43,7 +43,7 @@ services:
       POWERSYNC_CONFIG_PATH: /config/service.yaml
       PS_MONGO_URI: mongodb://mongo:27017/powersync
     networks:
-      - supabase_network_construculator-backend
+      - supabase_network_construculator-backend-e2e
     restart: unless-stopped
     healthcheck:
       test:
@@ -58,9 +58,9 @@ services:
       retries: 5
 
 networks:
-  supabase_network_construculator-backend:
+  supabase_network_construculator-backend-e2e:
     external: true
-    name: supabase_network_construculator-backend
+    name: supabase_network_construculator-backend-e2e
 
 volumes:
   mongo_storage:
--- a/supabase/config.toml
+++ b/supabase/config.toml
@@ -2,7 +2,7 @@
 # https://supabase.com/docs/guides/local-development/cli/config
 # A string used to distinguish different Supabase projects on the same host. Defaults to the
 # working directory name when running `supabase init`.
-project_id = "construculator-backend"
+project_id = "construculator-backend-e2e"
 
 [api]
 enabled = true
--- a/supabase/schemas/global_search/global_search/README.md
+++ b/supabase/schemas/global_search/global_search/README.md
@@ -49,7 +49,7 @@ supabase db reset
 
 # 2) Apply this module's schemas to the running local DB.
 for f in supabase/schemas/global_search/global_search/0?_*.sql; do
-  docker exec -i supabase_db_construculator-backend \
+  docker exec -i supabase_db_construculator-backend-e2e \
     psql -U postgres -d postgres -v ON_ERROR_STOP=1 < "$f"
 done
 
--- a/supabase/schemas/global_search/project_search_history/README.md
+++ b/supabase/schemas/global_search/project_search_history/README.md
@@ -129,7 +129,7 @@ supabase db reset
 
 # 2) Apply this module's schemas to the running local DB.
 for f in supabase/schemas/global_search/project_search_history/0?_*.sql; do
-  docker exec -i supabase_db_construculator-backend \
+  docker exec -i supabase_db_construculator-backend-e2e \
     psql -U postgres -d postgres -v ON_ERROR_STOP=1 < "$f"
 done
 


supabase/config.toml
```diff
@@ -2,7 +2,7 @@
 # https://supabase.com/docs/guides/local-development/cli/config
 # A string used to distinguish different Supabase projects on the same host. Defaults to the
 # working directory name when running `supabase init`.
-project_id = "construculator-backend"
+project_id = "construculator-backend-e2e"
```
ripplearcgit, line 5: 🔴 **The rename does not create a second project — dev and E2E still resolve to the same stack**

This is the finding that matters; the other two are downstream of it.

The PR's premise is that `project_id = "construculator-backend"` *"collided with whatever a developer's own local Supabase project defaults to from the same directory name."* That premise doesn't survive contact with how the two repos resolve the backend:

- `scripts/e2e/lib.sh` (app repo) sets `E2E_BACKEND_DIR="${E2E_BACKEND_DIR:-$(dirname "$E2E_APP_DIR")/construculator-backend}"` — the sibling checkout.
- This repo's own `README.md` §Getting Started tells a developer to `cd construculator-backend` then `npx supabase start` — *that same checkout*.

One directory holds one `config.toml`, which holds one `project_id`. Renaming it renames the single project both workflows already share.

The destructive commands confirm it: `E2E_DB_PROJECT` is used only in the confirmation *prompt strings* — it never selects a target. The real operations are `(cd "$E2E_BACKEND_DIR" && supabase db reset)` and `(cd "$E2E_BACKEND_DIR" && supabase stop --no-backup)`, which act on whatever `config.toml` is in that directory. After this PR that is `construculator-backend-e2e` — and so is the developer's dev stack, because it is the same file.

```mermaid
sequenceDiagram
    autonumber
    participant Dev as Developer
    participant E2E as scripts/e2e in app repo
    participant Repo as construculator-backend checkout
    participant CLI as Supabase CLI
    participant Docker as Docker

    rect rgba(139,92,246,0.15)
    Note over Dev,Docker: Analogy: one filing cabinet, two users. This PR only repaints the label on it.
    end

    Dev->>Repo: npx supabase start, per the README Getting Started
    Repo->>CLI: config.toml project_id = construculator-backend-e2e
    CLI->>Docker: creates supabase_db_construculator-backend-e2e
    Docker-->>Dev: ordinary dev data now lives in this volume

    E2E->>Repo: E2E_BACKEND_DIR defaults to this same sibling checkout
    Repo->>CLI: same config.toml, therefore the same project_id
    CLI->>Docker: supabase db reset and supabase stop --no-backup

    rect rgba(139,92,246,0.15)
    Note over CLI,Docker: Analogy: the cabinet is emptied, and the working papers were still inside it.
    end

    Docker-->>Dev: dev data destroyed, only the seeded fixtures come back
```

The one reading where the rename appears to work is a developer whose dev clone is still pinned to pre-merge `master` while the E2E clone is on this branch. That's a transient accident, not isolation: the moment the dev clone pulls `master`, both names converge again.

**What would actually deliver the isolation:** give the E2E harness its own backend checkout (or its own config), so there are genuinely two `config.toml` files — and move the ports too (see Nit 1), since two stacks that bind the same ports still can't run at once.

Worth flagging for sequencing: app#576 rewrites the confirmation prompts and docs to say only E2E data is at stake. Until this blocker is closed, that change would make the destructive scripts read as *safer* than they are.

powersync/.env.example
```diff
@@ -5,5 +5,5 @@
 PS_API_TOKEN=your-secure-api-token-here
 
 PS_PORT=8081
-PS_POSTGRES_URI=postgresql://postgres:postgres@supabase_db_construculator-backend:5432/postgres
-PS_JWKS_URI=http://supabase_kong_construculator-backend:8000/auth/v1/.well-known/jwks.json
+PS_POSTGRES_URI=postgresql://postgres:postgres@supabase_db_construculator-backend-e2e:5432/postgres
+PS_JWKS_URI=http://supabase_kong_construculator-backend-e2e:8000/auth/v1/.well-known/jwks.json
```
ripplearcgit, line 9: 🔴 **Every developer with an existing `powersync/.env` gets a broken stack and a misleading error**

`powersync/.env` is gitignored and generated once from this example file. Updating the example doesn't update anybody's existing file — and the generator explicitly refuses to: `e2e_ensure_powersync_env()` opens with `[ -f "$env_file" ] && return 0`.

So on every machine that has already run this stack, `PS_POSTGRES_URI` and `PS_JWKS_URI` keep pointing at `supabase_db_construculator-backend` and `supabase_kong_construculator-backend` — hostnames that no longer exist after the rename. Those values reach the service for real: `compose.yaml` declares `env_file: - .env` on the powersync service, and `service.yaml` resolves `uri: !env PS_POSTGRES_URI` and `jwks_uri: !env PS_JWKS_URI`.

The failure is neither fast nor legible. `docker compose up -d` succeeds — the network exists, the container starts. PowerSync then can't resolve the old hostname, its health endpoint never comes up, and `start_env.sh` polls for 120 seconds before dying with `PowerSync did not become ready within 120s`. Nothing in that message mentions a hostname, an `.env`, or a rename.

```mermaid
flowchart TD
    A["git pull brings in this PR"] --> B["powersync/.env already exists on disk"]
    B --> C["e2e_ensure_powersync_env sees the file and returns early"]
    C --> D["Stale PS_POSTGRES_URI still names supabase_db_construculator-backend"]
    D --> E["supabase start creates supabase_db_construculator-backend-e2e instead"]
    E --> F["docker compose up -d succeeds, the container does start"]
    F --> G["PowerSync cannot resolve the old hostname"]
    G --> H["start_env.sh polls the health endpoint for 120s"]
    H --> I["error: PowerSync did not become ready within 120s"]
    I --> J["The message names the timeout, never the stale hostname"]
    style D fill:#fce8e6,stroke:#d93025
    style J fill:#fce8e6,stroke:#d93025
```

**Fix (~6 lines here):** add a short "Migrating from the old project_id" block to `powersync/README.md` telling developers to delete `powersync/.env` and let it regenerate. Better still, make it self-healing on the app side so nobody has to read a README — reconcile stale hostnames in `e2e_ensure_powersync_env()` rather than returning early whenever the file merely exists.

ripplearcgit (changes requested):
## Review — CA-991: isolate the E2E project_id from ordinary local dev

**Verdict: Request changes** — 3 blocking, 4 nits.

The mechanical rename in this PR is clean, and I want to be precise about that before the blockers: a full-tree grep at `ad6a955` finds **zero stale Docker identifiers**. Fifteen occurrences carry the new `-e2e` suffix; the five remaining bare `construculator-backend` strings are all correct to leave alone (a `cd`, two directory-tree diagrams, the repo URL, and the npm package name). CI is `project_id`-agnostic (`supabase db start` / `supabase test db`), and `E2E_DB_PROJECT` in app#576 matches this PR's `project_id` character-for-character.

The problem is not that the rename is half-applied. It's that **renaming the shared `config.toml` doesn't split one stack into two** — so the isolation the PR sets out to deliver isn't delivered, and two migration hazards ship with it. Two blockers are inline on the lines they attach to; the third is below.

---

### 🔴 Blocker 3 — Pre-rename containers and volumes are orphaned, with no documented cleanup

Nothing here stops, removes, or renames the stack developers are already running. Two concrete consequences, both undocumented:

**Port conflict.** After pulling, `supabase stop` matches only `construculator-backend-e2e`, so the old containers keep running and keep holding 54321–54324. The next `supabase start` fails with `Bind for 0.0.0.0:54321 failed: port is already allocated`. Recovering requires knowing to hunt down containers under a name that no longer appears anywhere in the repo.

**Dev data goes invisible.** The new project creates a fresh, empty volume. The developer's existing local database is still on disk in `supabase_db_construculator-backend`, but nothing references it any more — it looks, from the developer's seat, exactly like the data loss this PR was written to prevent. It's recoverable, but only by someone who already knows what happened.

```mermaid
flowchart TD
    A["Developer already has the pre-rename stack up"] --> B["Containers named supabase_db_construculator-backend hold ports 54321 to 54324"]
    A --> C["Their local dev data sits in the old named volume"]
    B --> D["Merge renames project_id to construculator-backend-e2e"]
    C --> D
    D --> E["supabase stop now matches the new name only"]
    E --> F["Old containers keep running and keep the ports"]
    F --> G["supabase start fails: port 54321 already allocated"]
    D --> H["New project creates a fresh empty volume"]
    H --> I["Old volume is orphaned. Data is invisible, though not deleted"]
    G --> J["No cleanup step is documented anywhere in this PR"]
    I --> J
    style G fill:#fce8e6,stroke:#d93025
    style J fill:#fce8e6,stroke:#d93025
```

This is the cheapest of the three to close — a documentation gap, not a design one. A one-time migration note in `powersync/README.md` (and ideally the root `README.md`) covering: retire the old containers, delete `powersync/.env` so it regenerates, and run `docker compose -f powersync/compose.yaml down --volumes` so PowerSync's stale Mongo bucket storage goes too (see Nit 2).

---

*Nits, verified-good findings, and the full next-steps checklist follow in a separate comment.*

ripplearcgit (commented):
## Nits, verified-good, and next steps

Non-blocking. None of these attach cleanly to a changed line, so they're all here.

### 🟡 Nits

**1 — Ports are unchanged, so even a genuinely separate project couldn't coexist.** `supabase/config.toml` keeps 54321 (API), 54322 (db), 54323 (Studio), 54324 (Mailpit), and `.env.example` keeps `PS_PORT=8081`. A separate `project_id` gives you separate containers and volumes, but two stacks still can't be up at once if they bind the same ports — which is the whole point of having two. This is corroborating evidence for Blocker 1 rather than an independent bug: it shows the isolation was reasoned about at the naming layer only. Any real fix has to move the ports too, and then `E2E_SUPABASE_API_PORT` / `E2E_SUPABASE_DB_PORT` / `E2E_MAILPIT_PORT` / `E2E_POWERSYNC_PORT` in the app's `lib.sh` must follow, since that file's own comment says it only mirrors them.

**2 — PowerSync's MongoDB bucket storage survives the rename and describes the replaced database.** The `mongo_storage` volume belongs to the Compose project (`powersync_mongo_storage`), not the Supabase project, so the rename doesn't touch it. Afterwards Supabase hands PowerSync a brand-new database with a new replication slot while MongoDB still holds buckets describing the old one. The repo already knows this is a hazard — `reset_env.sh`'s header says the bucket storage *"still describe[s] the database that was just replaced"*, and it runs `compose down --volumes` for exactly that reason. The manual path in `powersync/README.md` — the path this PR's own unchecked test-plan box describes — has no equivalent step.

**3 — The unchecked test-plan box is the one that would have caught Blockers 2 and 3.** The only completed verification is the `docker compose config` parity check (`powersync/compose.yaml` parsed via PyYAML). That confirms the YAML is well-formed and internally consistent, which it is. It cannot detect a stale `.env`, an orphaned container, a port conflict, or a hostname that doesn't resolve, because none of those are properties of the file. The unchecked box — `supabase start` + `docker compose up -d`, confirming the containers/network are named `..._construculator-backend-e2e` — is precisely the step that surfaces all of them, and it's where your own machine would have hit the 120-second PowerSync timeout. Worth stating plainly rather than treating as routine: **for an infra-rename PR, the manual bring-up is the test.**

**4 — Nothing in this repo tells a reader the stack is now meant to be E2E-only.** The `-e2e` suffix changes what this project *means*, but the backend repo never says so. `README.md` §Getting Started still presents `npx supabase start` as the ordinary way to run the backend locally; `powersync/README.md` still reads as general local-dev documentation; the two `global_search` READMEs still describe `docker exec supabase_db_construculator-backend-e2e` as the normal way to apply a schema module by hand. A backend developer working only in this repo sees an `-e2e` suffix on their everyday containers with no explanation. This is Blocker 1 seen from the documentation side: if the stack really is E2E-only, this repo needs to say where ordinary dev work happens instead — and if it isn't, the suffix is misleading and app#576's docs are wrong.

<details>
<summary><b>Verified good — what was checked and passed</b></summary>

| # | Check | Result |
|---|---|---|
| 1 | Rename completeness | ✅ Full-tree grep at `ad6a955`: 15 occurrences carry `-e2e`; the 5 remaining bare strings are a `cd`, two tree diagrams, the repo URL, and the npm package name. No Makefile, no justfile, no script referencing container names, no committed `.env`. |
| 2 | Derived container/network names | ✅ The CLI composes `supabase_<service>_<project_id>`. Confirmed empirically against the pre-PR state, which has been working since #37. `construculator-backend-e2e` is a legal Compose project name. |
| 3 | CI impact | ✅ None. `database_tests.yml` (and the stray legacy copy at `.github/database_tests.yml`) run `supabase db start` / `supabase test db` — both `project_id`-agnostic. A fresh runner has no old stack, so Blockers 2 and 3 are local-machine-only. |
| 4 | Cross-repo consistency with app#576 | ✅ Exact. `E2E_DB_PROJECT` at #576's head (`ba062d2`) matches character-for-character. No other identifier crosses the repo boundary. |
| 5 | `external: true` network reference | ✅ Unchanged in kind — `master` already carries it on the old name. This PR changes which network is required, not the requirement. Failure mode is legible: Compose names the missing network exactly, and `start_env.sh` orders `supabase start` first. |

**Caveat on check 2:** Docker wasn't running on the review machine, so the derivation is verified by construction plus the working prior state, not by starting a live stack. That's the same gap Nit 3 is about.

</details>

<details>
<summary><b>Merge ordering — and the interaction with #51</b></summary>

**Ordering.** Because `E2E_DB_PROJECT` only feeds prompt text, this PR landing without app#576 does not break `start_env.sh` — the stack still comes up, the prompts just name the wrong project. The risk runs the other way: **app#576 must not merge without this PR**, or its prompts name a project that doesn't exist. Given Blocker 1, the safer sequencing is this PR alone as inert cleanup, with #576's warning removals held back entirely until the isolation is real.

**Sibling #51 (CA-749).** #51 seeds a GoTrue credential with a published password into the shared seeder set, and its safety argument leans on the stack being local and isolated. Three observations:

- **Scope** — this PR doesn't change what is local versus remote. It changes Docker naming inside a stack that was already local-only. #51's actual risk surface is untouched by this PR in either direction.
- **Premise** — the "isolated" half is not delivered even locally, per Blocker 1. The seeded published-password account lands in the same database a developer uses for ordinary work.
- **No ordering constraint between the two.** This PR opens no isolation window for #51 to depend on. If #51's placement is acceptable today, it's equally acceptable after this; if it isn't, this doesn't make it so.

Worth noting: #51's own test plan used the correct recipe — a distinct `project_id` **and** distinct ports (`ca749probe`, 556xx), in a copy of the repo. That's exactly the shape Blocker 1 and Nit 1 are asking for here.

</details>

### Suggested next steps

- [ ] `supabase/config.toml:5` — the rename doesn't split dev from E2E; both resolve to this one checkout. Give the E2E harness its own checkout or config 🔴
- [ ] `powersync/.env.example:8-9` — existing `powersync/.env` files keep the old hostnames and are never regenerated. Document deleting the file, or make `e2e_ensure_powersync_env()` self-healing 🔴
- [ ] `powersync/README.md` — add a "Migrating from the old project_id" block: remove pre-rename containers (they hold 54321–54324), delete the stale `.env`, clear the Mongo volume 🔴
- [ ] `supabase/config.toml` — move the ports (54321–54324, `PS_PORT=8081`) if the two stacks are ever meant to run concurrently; mirror into the app's `lib.sh` 🟡
- [ ] `powersync/compose.yaml` — fold `docker compose … down --volumes` into the migration note so stale Mongo bucket storage is cleared 🟡
- [ ] Run the unchecked manual test-plan step on a machine that already has the old stack — it's the step that reproduces Blockers 2 and 3 🟡
- [ ] `powersync/README.md` / root `README.md` — state what the `-e2e` suffix means and where ordinary backend dev work is supposed to run 🟡
- [ ] Coordination — app#576 must not merge before this PR, and its warning removals shouldn't merge at all until Blocker 1 is closed

supabase/config.toml
ayanasamuel8: Agreed the rename alone doesn't deliver isolation — thanks for tracing `E2E_BACKEND_DIR` and the README's `cd construculator-backend` down to the same checkout, that's the crux of it.

What I've done in 086a3b8, scoped to what's fixable from this repo alone:

- **Moved every local port** (`supabase/config.toml`: API/DB/shadow/pooler/Studio/Inbucket/Analytics/edge-runtime-inspector, `powersync/.env.example` + `compose.yaml`: `PS_PORT`) off the CLI's `543xx`/`8081`/`8083` defaults, shifted by `+1000`. This is Nit 1's fix folded in here since it's the same root cause. Verified locally: brought up the `-e2e` Supabase + PowerSync stack **while a separate pre-existing non-`-e2e` `construculator-backend` stack was already running** on the same machine — both stayed up concurrently with zero port conflicts, PowerSync `probes/liveness` returned `ready:true` and replicated seeded rows out of the `-e2e` database.
- **Added an explicit "What this checkout is for" note** to `powersync/README.md`, plus a matching note in the root `README.md`'s Getting Started, stating in plain terms that this `project_id` is for `construculator-app`'s E2E harness, and that real isolation from everyday dev work requires the harness to target a *separate* checkout — calling out by name that `E2E_BACKEND_DIR`'s sibling-directory default will otherwise resolve to the same checkout a developer already uses.

What I haven't done, because it isn't achievable from this repo: forcing `E2E_BACKEND_DIR` to actually point at a dedicated checkout. Supabase's CLI ties `project_id`/migrations/schemas to one `config.toml` per checkout tree — there's no clean way to fork that within a single clone (a `--workdir`-based second config would mean duplicating `supabase/migrations`, `supabase/schemas`, etc., which trades this bug for a drift bug). The real fix is `construculator-app` pointing `E2E_BACKEND_DIR` at a second, dedicated clone, which is app#576's territory (or a follow-up there).

So: two of the three concrete failure modes you raised here (port collision, "which checkout is this for") are closed. The "same checkout" risk is documented but not structurally prevented — I'm leaving this thread open rather than resolving it, since the finding that started it still stands until app#576 (or a follow-up) makes the checkout separation real. Flagging to the team to make sure that's tracked before app#576 removes its warning language, per your sequencing note.

powersync/.env.example
ayanasamuel8: Fixed in 086a3b8 with the ~6-line doc block you suggested, plus a bit more since it turned out to be the natural place to also cover Blocker 3 / Nit 2.

Added a "Migrating from the old project_id" section to `powersync/README.md` that has a developer with an existing stack:
1. `docker compose -f powersync/compose.yaml down --volumes` — removes the pre-rename PowerSync containers *and* the Mongo bucket-storage volume (Nit 2 — it still describes the pre-rename database and would otherwise silently survive).
2. `npx supabase stop --project-id construculator-backend --no-backup` (with a raw `docker rm`/`docker volume rm` fallback) — retires the pre-rename Supabase containers/volumes that `supabase stop` (no args) can no longer see, since it now only matches `construculator-backend-e2e` (Blocker 3).
3. `rm -f powersync/.env && cp powersync/.env.example powersync/.env` — regenerates the gitignored `.env` so the stale `supabase_db_construculator-backend` / `supabase_kong_construculator-backend` hostnames you flagged here are gone, since nothing in this repo will overwrite it automatically.

I didn't do the "better still" app-side self-healing (`e2e_ensure_powersync_env()` reconciling stale hostnames instead of returning early) — that function lives in `construculator-app`, not here, so it's out of reach from this PR. Worth a follow-up issue there if you want the belt-and-suspenders version; the doc fix at least makes the failure mode legible and fixable in one copy-paste.

Resolving this thread since the requested fix is in.

ayanasamuel8:
Replying to the non-inline findings from the two review comments (Blocker 3, and the nits/next-steps), since GitHub won't let me thread a reply directly under them. Pushed in 086a3b8.

**🔴 Blocker 3 — orphaned pre-rename containers/volumes, no documented cleanup.** Fixed: added the "Migrating from the old project_id" section to `powersync/README.md` (mirrors what I replied on the `.env.example` thread) — retire the old Supabase containers/volumes with `supabase stop --project-id construculator-backend --no-backup`, clear PowerSync's stale Mongo bucket storage with `docker compose down --volumes`, and regenerate `powersync/.env`.

**🟡 Nit 1 — ports unchanged.** Fixed: moved every local port `+1000` off the Supabase CLI/PowerSync defaults (`supabase/config.toml`, `powersync/.env.example`, `powersync/compose.yaml`'s fallback). Verified two stacks can now actually run concurrently (details in the reply on the `config.toml` thread). I did **not** touch `construculator-app`'s `lib.sh` — that mirroring has to happen in app#576 since it's a different repo; flagging it there is on the coordination list below.

**🟡 Nit 2 — Mongo bucket storage.** Fixed: folded `docker compose down --volumes` into the migration doc, see Blocker 3 above.

**🟡 Nit 3 — unchecked test-plan box.** Ran it: `supabase start` + `docker compose -f powersync/compose.yaml up -d`, confirmed `-e2e`-suffixed containers/network on the new ports, PowerSync `probes/liveness` → `{"ready":true,"started":true}`, and live replication of seeded rows. Did this with a separate pre-existing non-`-e2e` stack already running, so it doubled as a concurrency check. Checked off in the PR description, with the details recorded there.

**🟡 Nit 4 — nothing says the stack is E2E-only.** Fixed: added a "What this checkout is for" section to `powersync/README.md` and a note in the root `README.md` Getting Started, both stating the `-e2e` project_id is for the E2E harness and that sharing a checkout between that and daily dev defeats the isolation.

**Coordination / not done here (needs app#576 or a follow-up, not fixable from this repo alone):**
- The structural fix for Blocker 1 — E2E harness targeting a dedicated second checkout, since `E2E_BACKEND_DIR`'s sibling-directory default can otherwise resolve to the same checkout as daily dev. Replied on that thread with the reasoning; left it unresolved.
- Mirroring the new ports into `construculator-app`'s `lib.sh` (`E2E_SUPABASE_API_PORT` etc.) — app-repo change.
- The optional app-side self-healing for `e2e_ensure_powersync_env()` — app-repo function, the doc-based fix here covers the same failure mode.
- Merge ordering: agreed with your note that app#576 shouldn't merge first, and that its warning-removal shouldn't land until the checkout-separation piece is real. Nothing in this PR's scope changes that.

powersync/compose.yaml
```diff
@@ -58,9 +58,9 @@ services:
...
 
 networks:
-  supabase_network_construculator-backend:
+  supabase_network_construculator-backend-e2e:
     external: true
-    name: supabase_network_construculator-backend
+    name: supabase_network_construculator-backend-e2e
 
 volumes:
   mongo_storage:
```
ripplearcgit, line 68: 🔴 **Blocker 2 — PowerSync's own Docker identities were never namespaced, so the two-checkout remedy the new docs prescribe does not actually isolate PowerSync**

This PR namespaces the Supabase half of the stack thoroughly. The PowerSync half was left entirely global, and nothing in either round noticed. Three separate mechanisms are involved and all three point the same way:

1. **Three hardcoded `container_name` values** — `powersync_mongo` (L6), `powersync_mongo_init` (L16), `powersync` (L31). A `container_name` opts a service out of Compose's per-project naming: the name becomes global to the Docker daemon. Two checkouts cannot both have a container called `powersync`.
2. **An identical Compose project name in both checkouts** — the file has no top-level `name:` key, so Compose derives the project from the directory holding the compose file. That is `powersync` whether you run the migration doc's `docker compose -f powersync/compose.yaml …` from the repo root, or the app's `e2e_powersync_compose()`, which passes `--project-directory "$E2E_BACKEND_DIR/powersync"`. Same name, both times, in every checkout.
3. **Therefore one shared volume** — `mongo_storage` (this line) resolves to `powersync_mongo_storage` in both checkouts. One disk, two stacks.

Put together, in the exact setup `powersync/README.md` now tells developers to build: the second `docker compose up -d` either fails with `Conflict. The container name "/powersync" is already in use`, or — because Compose sees a project of the same name — treats the first checkout's containers as its own and recreates them onto the E2E network, pointed at the E2E database. The dev checkout's PowerSync silently becomes the E2E one.

The sharper consequence is destructive, and it is the same failure this PR was written to prevent. `reset_env.sh` runs `e2e_powersync_compose down --volumes` before every reset. In a two-checkout setup that deletes `powersync_mongo_storage` — which is also the dev checkout's bucket storage. So a developer who follows this PR's new documentation to the letter still loses local state to the E2E scripts, just on the PowerSync side rather than the Postgres side.

```mermaid
flowchart TD
    subgraph Dev["Dev checkout (any project_id)"]
      D1["supabase_db_my-dev on 543xx"]
      D2["docker compose -f powersync/compose.yaml up -d"]
    end
    subgraph E2E["E2E checkout (construculator-backend-e2e)"]
      E1["supabase_db_construculator-backend-e2e on 553xx"]
      E2["docker compose -f powersync/compose.yaml up -d"]
    end
    D1 -.-> S1
    E1 -.-> S1
    S1["Supabase halves are now genuinely separate. This PR fixed that."]
    D2 --> P["Compose project name is derived from the powersync directory. Identical in both checkouts."]
    E2 --> P
    P --> Q["container_name powersync, powersync_mongo, powersync_mongo_init are hardcoded, no project_id suffix"]
    P --> R["volume resolves to powersync_mongo_storage in both"]
    Q --> T["Second up -d conflicts on the container name, or recreates the first checkout containers onto the e2e network"]
    R --> U["reset_env.sh runs compose down --volumes and deletes the dev checkout bucket storage too"]
    style S1 fill:#e6f4ea,stroke:#34a853
    style T fill:#fce8e6,stroke:#d93025
    style U fill:#fce8e6,stroke:#d93025
```

**Fix — small, and it belongs in this PR.** Add a top-level project name and drop the three pins, so Compose namespaces everything the same way the Supabase CLI now does:

```diff
+name: construculator-backend-e2e-powersync   # new top-level key
+
 services:
   mongo:
-    container_name: powersync_mongo
   mongo-rs-init:
-    container_name: powersync_mongo_init
   powersync:
-    container_name: powersync
```

Two consequences to carry across if you do this: the container names in `powersync/README.md`'s troubleshooting section change shape, and the app's `e2e_powersync_compose()` keeps working unchanged because it addresses the stack through Compose rather than by container name.

Also note this reframes round-1 Nit 3: the manual concurrency test you ran and reported can only have exercised Supabase-against-Supabase, since two PowerSync stacks cannot coexist as the file stands. That is **not** a claim of dishonesty — it is a gap the test could not have surfaced.

README.md
```diff
@@ -60,6 +60,15 @@ Database backend for **Construculator** — a construction cost estimation platf
...
 ## Getting Started
 
+> **Note:** `supabase/config.toml`'s `project_id` is `construculator-backend-e2e` — this checkout's
+> Supabase stack is named for `construculator-app`'s E2E test harness, whose scripts run
+> destructive commands (`supabase db reset`, `supabase stop --no-backup`) against whatever
+> checkout they're pointed at. If you're also using this checkout as the E2E harness's target
+> (`construculator-app`'s `E2E_BACKEND_DIR`, which defaults to a sibling directory named
+> `construculator-backend`), be aware you're sharing one database between your manual work below
+> and those destructive scripts. For a stack that's genuinely yours, either keep the E2E harness
+> pointed at a separate clone, or use a different `project_id` in your own checkout.
```
ripplearcgit: 🟨 **Major 1 — The second remedy offered here does not work, and recreates the original bug**

The note closes with two options presented as equivalent: *"either keep the E2E harness pointed at a separate clone, or use a different `project_id` in your own checkout."* Only the first one is real.

If the harness is still pointed at this checkout, changing `project_id` changes nothing about what gets destroyed. `reset_env.sh` runs `(cd "$E2E_BACKEND_DIR" && supabase db reset)`. That command reads whatever `config.toml` is in that directory and rebuilds whatever project it names. Rename the project to `my-dev` and the script now destroys `my-dev` — which is the developer's own stack, by construction, since renaming it was their idea. The advice reproduces the precise failure this PR exists to prevent, and does so on a reader who was trying to follow the safety note.

Two smaller problems ride along. `E2E_DB_PROJECT` in `lib.sh` is a hardcoded literal, so the confirmation prompt would then name a project that does not exist — the one place the developer might catch the mistake now shows them a reassuring, wrong string. And `supabase/config.toml` is a tracked file, so the change leaves the checkout permanently dirty and one `git commit -a` away from landing on master.

```mermaid
flowchart TD
    A["Root README: for a stack that is genuinely yours, either A or B"]
    A --> B["Option A: point E2E_BACKEND_DIR at a separate clone"]
    A --> C["Option B: use a different project_id in your own checkout"]
    B --> D["Two config.toml files. reset_env.sh acts on the other one. Isolation is real."]
    C --> E["Still one checkout. E2E_BACKEND_DIR still resolves here."]
    E --> F["reset_env.sh does cd E2E_BACKEND_DIR then supabase db reset"]
    F --> G["It acts on whatever config.toml is in that directory, whatever it is named"]
    G --> H["The renamed dev project is the one destroyed. Original bug recreated."]
    G --> I["E2E_DB_PROJECT prompt now names a project that does not exist"]
    C --> J["supabase/config.toml is tracked, so the checkout stays permanently dirty"]
    style D fill:#e6f4ea,stroke:#34a853
    style H fill:#fce8e6,stroke:#d93025
    style I fill:#fef9c3,stroke:#ca8a04
    style J fill:#fef9c3,stroke:#ca8a04
```

**Fix:** delete the `or…` clause. `powersync/README.md` already gets this right — it says plainly "use two checkouts" and explains why a rename alone does nothing when both workflows share a directory. The two documents currently disagree, and the root README is the one a new developer reads first.

powersync/README.md
```diff
@@ -2,6 +2,57 @@
...
+# 1. Stop and remove the pre-rename PowerSync stack, including the Mongo bucket-storage volume.
+#    Bucket storage describes replication state tied to the old database, so it must go too —
+#    reusing it against the renamed database would make PowerSync sync from stale state.
+docker compose -f powersync/compose.yaml down --volumes
+
+# 2. Stop and remove the pre-rename Supabase stack, including its data volumes. This targets the
+#    old project_id explicitly, since `supabase stop` (no args) only matches the project_id in
+#    the current config.toml, which is now construculator-backend-e2e. If you want to keep the
+#    old dev data instead of discarding it, drop --no-backup and back up the volume first.
+npx supabase stop --project-id construculator-backend --no-backup
```
ripplearcgit: 🟨 **Major 3 — The migration recipe's copy-paste default destroys the pre-rename dev database**

Step 2's runnable line is `npx supabase stop --project-id construculator-backend --no-backup`. The CLI's own help for that flag is unambiguous: *"Deletes all data volumes after stopping."* The volume in question is precisely the one round-1 Blocker 3 identified as holding "the developer's existing local database" — real work, not fixtures.

The preserve-it advice does exist, but it is a comment three lines above the command, inside a fenced block presented as "Clean up the old stack once, from the repo root" with three numbered copy-paste steps. Readers copy the block. For a PR whose entire premise is that a destructive command was being aimed at real dev data, shipping a migration recipe whose default aims a destructive command at real dev data is the wrong way round.

**Fix — one line, no new content.** Invert the default: make `npx supabase stop --project-id construculator-backend` the copy-paste line (modern CLI versions dump the database before stopping, so this is the recoverable path), and demote `--no-backup` to a follow-on comment for readers who have confirmed they want the volumes gone.

ripplearcgit (changes requested):
## Round 2 re-review — 🔴 CHANGES REQUESTED

**2 blockers · 3 majors · 4 nits · 7 verified good** — reviewed against `086a3b8`, not against the PR body.

Round 1 was answered seriously. Five of its seven findings are genuinely closed, the mechanical work is clean, and the new documentation is honest about what the rename does and does not buy — including the part you could not fix from this repo. That is the right disposition, and this review should not read as a rejection of it.

What blocks the merge is **not** the acknowledged gap. It is that the round-1 fix moved an interface without moving its consumer, and that the half of the stack nobody has looked at yet — PowerSync's own Docker identities — reproduces the original collision inside the very two-checkout setup the new docs prescribe. The first is a coordination fix in another repo; the second is roughly four lines in `powersync/compose.yaml` and belongs here.

On the framing question — does this ship a false sense of isolation? **Not from this repo's side.** Both new README sections say plainly that real isolation needs a second checkout, and they name the mechanism. The destructive risk is *renamed, not removed*, and this PR's docs say so. But the *pair* is not honest yet: `construculator-app`#576 still carries a `lib.sh` comment asserting that the scripts operate on "their own" containers and "not a developer's ordinary dev stack," and it removes the warning language from all three scripts. That comment is false under the default configuration, and it sits in the file developers actually read when running the destructive commands. Fix app#576's comment in the same breath as its ports, and the pair becomes truthful.

### Architectural overview

```mermaid
sequenceDiagram
    autonumber
    participant Dev as Developer
    participant Cfg as The one config.toml
    participant CLI as Supabase CLI
    participant Docker as Docker
    participant E2E as app scripts/e2e

    rect rgba(139,92,246,0.15)
    Note over Dev,E2E: Analogy: one shared kitchen. This PR renames it and moves it to another floor.
    end

    Dev->>Cfg: npx supabase start in this checkout
    Cfg->>CLI: project_id construculator-backend-e2e, ports 55321 to 55329
    CLI->>Docker: creates supabase_db_construculator-backend-e2e
    Note over Docker: New names plus new ports no longer collide with a 543xx dev stack

    E2E->>Cfg: E2E_BACKEND_DIR still defaults to this same checkout
    Cfg->>CLI: same file, therefore the same project and the same database

    rect rgba(139,92,246,0.15)
    Note over CLI,Docker: Analogy: the cleaner still holds the address of the only kitchen there is.
    end

    CLI->>Docker: supabase db reset destroys the developer data too
    Docker-->>E2E: start_env.sh polls 54321 and 8081, which this PR just retired
    E2E--xDev: PowerSync did not become ready within 120s
```

> ⚠️ The direction is right and the in-repo execution is clean, but the change is not self-contained: it moves an interface that a second repository reads from without moving the other side, and it namespaces only half of the stack.
>
> **In plain English:** think of a shared kitchen. Two cooks were using the same one, and every night a cleaner came in and threw out everything on the counters. This change gives the kitchen a new name and moves it to a different floor, so a second kitchen could now exist without the two clashing. But nobody has actually built the second kitchen. The cleaner still has the address of the only one there is. And the pantry both kitchens draw from was never renamed at all, so it is still one pantry.
>
> Maps to: the "kitchen" is this checkout's `supabase/config.toml`; the "new name" is `project_id = construculator-backend-e2e`; the "different floor" is the +1000 port shift; the "cleaner" is `reset_env.sh` / `stop_env.sh --purge`; the "address" is `E2E_BACKEND_DIR`, still defaulting to this same directory; the "pantry" is `powersync/compose.yaml`'s hardcoded `container_name` values and its shared `powersync_mongo_storage` volume.

---

## 🔴 Blocker 1 — The ports moved; their only consumer did not. Merging this breaks the E2E harness.

*(out of diff — the file that needs the fix is `scripts/e2e/lib.sh` in **`ripplearc/construculator-app` PR #576**, so this cannot be an inline comment here.)*

This is a new blocker, created by the round-1 fix rather than left over from round 1. The port shift is correct and complete inside this repo. But these port numbers are an **interface**: `construculator-app`'s `scripts/e2e/lib.sh` declares its own copy of them and says, in its own comment, that the copy must be kept in step.

At app#576's head `ba062d2` — unchanged since round 1 — that file still reads:

```bash
# Read-only mirrors of the ports the backend already binds: ...
# To actually move a port, change it in the backend repository and update
# the matching value here.
E2E_SUPABASE_API_PORT=54321      # this PR moves it to 55321
E2E_SUPABASE_DB_PORT=54322       # this PR moves it to 55322
E2E_MAILPIT_PORT=54324           # this PR moves it to 55324
E2E_POWERSYNC_PORT=8081          # this PR moves it to 9081
```

Those four values are not decorative. `start_env.sh` polls `http://127.0.0.1:${E2E_SUPABASE_API_PORT}/auth/v1/health`, then `${E2E_MAILPIT_PORT}/api/v1/messages`, then `${E2E_POWERSYNC_PORT}/api/health`; `reset_env.sh` polls the PowerSync one again after the reset. After this PR nothing is listening on any of them, so `e2e_wait_for_http` loops for 120 seconds and then hard-exits with `error: Supabase auth did not become ready within 120s` — a message that never mentions a port.

**Blast radius, stated fairly:** `scripts/e2e/` does not exist on the app's `main` yet, so this does not break anything already merged. It breaks the unmerged #564 → #576 stack. That makes it a merge-gate problem rather than a production one — but it is a hard gate in *both* directions: round 1 established that app#576 must not merge before this PR, and this finding establishes that this PR must not merge before app#576 carries the new port numbers. They now have to land together.

```mermaid
flowchart TD
    A["This PR moves every backend port by plus 1000"] --> B["Supabase API 55321, DB 55322, Mailpit 55324"]
    A --> C["PowerSync host port 9081"]
    B --> D["app#576 head ba062d2 scripts/e2e/lib.sh is untouched"]
    C --> D
    D --> E["E2E_SUPABASE_API_PORT still 54321"]
    D --> F["E2E_MAILPIT_PORT still 54324"]
    D --> G["E2E_POWERSYNC_PORT still 8081"]
    E --> H["start_env.sh polls 127.0.0.1:54321/auth/v1/health"]
    F --> H
    G --> I["start_env.sh and reset_env.sh poll 127.0.0.1:8081/api/health"]
    H --> J["Nothing is listening. e2e_wait_for_http loops for 120s"]
    I --> J
    J --> K["error: Supabase auth did not become ready within 120s"]
    K --> L["lib.sh own comment: to move a port, update the matching value here. Not done."]
    style J fill:#fce8e6,stroke:#d93025
    style K fill:#fce8e6,stroke:#d93025
    style L fill:#fce8e6,stroke:#d93025
```

**Fix:** push the four new values into **`construculator-app`#576's `scripts/e2e/lib.sh`** and re-run its own unchecked manual box, then merge the two together. You flagged this as a coordination item in the PR thread, which is right — but flagging it does not unblock it, because until it is done the two branches describe different machines. Worth noting the asymmetry: `E2E_DB_PROJECT` only ever feeds a prompt string, so a mismatch there is cosmetic; a port mismatch is fatal.

---

## 🟨 Major 2 — The app's `construculator_dev` environment group still points `SUPABASE_URL` at 54321

*(out of diff — lives in **`ripplearc/construculator-app`**, at `docs/Testing/CI-Scripts.md:304` plus the Codemagic environment group itself.)*

This is the second consumer of the moved ports, and it is a different one from Blocker 1. `construculator-app/docs/Testing/CI-Scripts.md:304` documents the Codemagic `construculator_dev` environment group as containing `SUPABASE_URL=http://localhost:54321`. That value is not documentation-only: `scripts/ci/generate_env_file.sh` writes it into `assets/env/.env`, and `lib/libraries/supabase/supabase_wrapper_impl.dart` reads it at runtime and refuses to initialise without it.

After this PR the local backend's API is on **55321**, so a dev-flavour build pointed at the local stack connects to nothing. Two things need updating and neither is in app#576's file list: the doc line, and the actual value stored in the Codemagic environment group. Flagging as major rather than blocking because it affects a local dev-build convenience path rather than CI or the E2E harness — but it is exactly the kind of drift that surfaces a week later as "the app won't log in on my machine."

---

🔴 **Blocker 2** and 🟨 **Major 1** / 🟨 **Major 3** are posted as inline comments on `powersync/compose.yaml`, `README.md` and `powersync/README.md`.

The 4 nits, the risk map, the round-1 verification table, the verified-good checks and a suggested-next-steps checklist follow in a second comment on this review.

ripplearcgit (commented):
## Round 2 re-review — nits, verification record and next steps

Continues the `REQUEST_CHANGES` review above. Nothing here blocks the merge.

---

### ⚫ Nit 1 — `cp .env.example .env` defeats the token generator it is meant to hand off to

`powersync/README.md:45` — step 3's `rm -f powersync/.env` is exactly right. The `cp` on the next line (`:46`) undoes half of it. The app's generator is guarded by *existence*, not by content:

```bash
e2e_ensure_powersync_env() {
  local env_file="$E2E_BACKEND_DIR/powersync/.env"
  [ -f "$env_file" ] && return 0                    # <-- the cp makes this true
  sed "s|^PS_API_TOKEN=.*|PS_API_TOKEN=$(openssl rand -hex 32)|" \
    "$E2E_BACKEND_DIR/powersync/.env.example" >"$env_file"
}
```

So after the `cp`, `start_env.sh` sees a file and returns early, and PowerSync's admin API token stays the literal placeholder `your-secure-api-token-here` from the committed example unless the reader also acts on the trailing comment. Low stakes on a local-only stack — client sync authenticates through JWKS, not this token — but it is a published credential guarding an admin surface, and it is avoidable. Stop the step at `rm -f powersync/.env` and say that the next `start_env.sh` regenerates it with a fresh token.

### ⚫ Nit 2 — Two commented-out ports were missed by the +1000 shift

`supabase/config.toml:88-89` still read `# smtp_port = 54325` and `# pop3_port = 54326`. They are inert while commented, so this is not a bug today. But they sit two lines below the port this PR *did* shift, inside the same `[inbucket]` block, and the comment directly above invites a developer to uncomment them. Doing so lands them back on the CLI's default numbers — the exact collision this PR exists to remove, reintroduced by following the file's own instructions. Shift them to `55325` / `55326` so the block is internally consistent.

These two lines are the only stale port references anywhere in the tree. A full-tree grep for `5432[0-9]`, `8081` and `8083` at `086a3b8` returns nothing else except the new explanatory comment on line 10, which correctly cites 54321 as the old value.

### ⚫ Nit 3 — 55320–55329 sits inside the operating system's ephemeral port range

The new block runs 55320–55329. On this machine `sysctl net.inet.ip.portrange.first` returns `49152`, so every one of those numbers is inside the range macOS hands out at random to ordinary outbound connections; Linux's default range (32768–60999) covers them too. The CLI's 543xx defaults are deliberately below that boundary.

The consequence is low-probability and high-annoyance: occasionally a browser tab or a background process will already hold, say, 55322, and `supabase start` fails with `port is already allocated` for no reproducible reason. Since the shift is arbitrary anyway, a block below 49152 — `44320–44329`, say — buys the same isolation with none of this. Not worth blocking on; worth changing while the numbers are still cheap to change.

### ⚫ Nit 4 — Merge conflict with sibling #51, and the risky half of it is invisible

#51 (*CA-749: seed the auth credential for the sample user*) rewrites the same `README.md` region. Its hunk **deletes** the entire `### Linking the Test User to Supabase Auth` section — including the `http://localhost:54323` line this PR is bumping to `55323` — and replaces it with a new `### Signing In as the Seeded Test User` section that contains no port references at all. Its context lines also reach up into the "Viewing OTP Codes" list this PR touches.

Git will conflict on the overlap, which is fine — conflicts get resolved. The quieter risk is the part that will *not* conflict: #51's replacement text has no ports in it, so if #51 lands second and the resolver simply takes #51's version of that section, the port fix is silently dropped rather than visibly clashing.

**Recommended merge order: land #51 first, then #52 drops its README hunk for that section.** #51 *deletes* the very lines #52 edits, so resolving the conflict in #52's favour would silently resurrect the manual Studio linking procedure that round 1 blocked on. Whoever merges second should re-read the whole *Local Services & Ports → Signing In* stretch for surviving 543xx references, not just clear the conflict markers.

---

<details>
<summary><b>Risk map — 8 files</b></summary>

| File | Δ | Verdict |
|---|---|---|
| `supabase/config.toml` | +11 −9 | needs attention |
| `powersync/compose.yaml` | +6 −6 | needs attention |
| `README.md` | +23 −11 | worth a look |
| `powersync/README.md` | +61 −10 | worth a look |
| `powersync/.env.example` | +3 −3 | safe |
| `docs/PowerSync.md` | +1 −1 | safe |
| `supabase/schemas/global_search/global_search/README.md` | +1 −1 | safe |
| `supabase/schemas/global_search/project_search_history/README.md` | +1 −1 | safe |

</details>

<details>
<summary><b>Round-1 findings — independently re-verified against <code>086a3b8</code> (5 of 7 genuinely closed)</b></summary>

Round 1 (2026-08-22, `CHANGES_REQUESTED`) raised 3 blockers and 4 nits at `ad6a955`. Every claim below was checked against the code on the branch, not against the PR body.

| Round-1 finding | Status | What the code actually shows |
|---|---|---|
| 🔴 **B1** — The rename does not create a second project; dev and E2E still resolve to the same stack | **Documented only** | Unchanged in substance, and you say so openly. `E2E_BACKEND_DIR` at app#576 head `ba062d2` still defaults to `$(dirname "$E2E_APP_DIR")/construculator-backend`; the root README still says `cd construculator-backend` then `npx supabase start`. One directory, one `config.toml`, one project. `reset_env.sh` still runs `cd "$E2E_BACKEND_DIR" && supabase db reset`, which acts on whatever config is there. **The destructive risk is renamed, not removed.** Two README sections now say so — but one offers a remedy that does not work (Major 1), and app#576's `lib.sh` still asserts the opposite in a code comment (Blocker 1). |
| 🔴 **B2** — Stale gitignored `powersync/.env` keeps pre-rename hostnames | **Fixed** | `powersync/README.md` step 3 covers it explicitly, naming both the stale hostname (`supabase_db_construculator-backend`) and the stale `PS_PORT`, and explaining why the file never self-updates. The app-side self-healing was correctly declined as out-of-repo. One residual nit on the `cp` line (Nit 1). |
| 🔴 **B3** — Pre-rename containers and volumes orphaned, no documented cleanup | **Fixed** | New "Migrating from the old project_id" section, steps 1–2. I verified `supabase stop --project-id <id> --no-backup` is a real flag against a live CLI, and that the raw `docker ps` / `docker volume ls` fallbacks parse and match the right names — including `xargs -r`, which does work on macOS. The recipe's destructive default is a separate finding (Major 3). |
| 🟨 **N1** — Ports unchanged, so two stacks still could not coexist | **Fixed here, broken across the repo boundary** | Inside this repo the shift is exhaustive: a full-tree grep at `086a3b8` for `5432[0-9]`, `8081` and `8083` returns nothing but one explanatory comment and two commented-out lines (Nit 2). But `lib.sh`'s own instruction — *"To actually move a port, change it in the backend repository and update the matching value here"* — was not followed. See Blocker 1. |
| 🟨 **N2** — PowerSync's Mongo bucket storage survives the rename | **Fixed** | Step 1 of the migration block is `docker compose -f powersync/compose.yaml down --volumes`, with a comment explaining that bucket storage describes the replaced database. Correct. |
| 🟨 **N3** — The unchecked test-plan box is the one that would have caught B2 and B3 | **Claimed, partly contradicted** | The box is now checked with a specific, plausible account (both stacks up at once, liveness probe returning `ready:true`, live replication). I cannot re-run a manual bring-up. But the run described can only have exercised Supabase against Supabase — three hardcoded `container_name` values in `compose.yaml` make two concurrent PowerSync stacks impossible, so the concurrency claim does not extend as far as the PR implies. See Blocker 2. |
| 🟨 **N4** — Nothing tells a reader the stack is now E2E-only | **Fixed** | New "What this checkout is for" section in `powersync/README.md` and a note at the top of the root README's Getting Started. Both are clear and honest about the limitation. The root one carries a bad second remedy (Major 1). |

**Net:** 5 of 7 round-1 findings are genuinely closed. B1 remains open by your own admission and by mine. N1 is fixed on this side of the boundary and has opened a new one. The re-review's two blockers are both *consequences* of the R1 fix rather than leftovers from it.

</details>

<details>
<summary><b>🟢 Verified good — what I checked at <code>086a3b8</code> and found correct</b></summary>

| Check | Result |
|---|---|
| Rename completeness | ✅ Full-tree grep for `construculator-backend`. Every derived Docker identifier carries `-e2e`. The remaining bare strings are all correct to leave: the npm package name in `package-lock.json`, a `cd` in the README, two directory-tree diagrams, the repo URL, and five deliberate pre-rename references inside the new migration section. Nothing was missed. |
| Port-shift completeness | ✅ Full-tree grep for `5432[0-9]`, `8081`, `8083`. Zero hits outside `config.toml`'s new explanatory comment and the two commented-out lines in Nit 2. No script, Makefile, workflow, or doc kept an old number. |
| `project_id` agreement with app#576 | ✅ Exact. `E2E_DB_PROJECT="construculator-backend-e2e"` at `ba062d2` matches this PR's `config.toml` character-for-character. (The ports do not — Blocker 1.) |
| CI impact | ✅ None. `.github/workflows/database_tests.yml` pins `supabase/setup-cli@v1` at 2.106.0 and runs `supabase db start` / `supabase test db` — both agnostic to `project_id` and to published host ports. The stray legacy copy at `.github/database_tests.yml` is a 267-byte fragment with no jobs and is equally unaffected. A fresh runner has no pre-rename stack, so the migration hazards are local-machine-only. |
| Migration commands | ✅ All four verified against live tooling: the `--project-id` flag is real (confirmed in `supabase stop --help`); the `docker ps -a --filter "name=supabase_.*_construculator-backend$"` fallback matches the pre-rename containers and, thanks to the `$` anchor, does not sweep up the new `-e2e` ones; `xargs -r` works on macOS (I ran it — a GNU-ism BSD `xargs` historically lacked); and step 1 correctly targets the pre-rename PowerSync stack even when run from the renamed checkout. |
| Internal vs. published ports | ✅ Correctly distinguished throughout. `service.yaml`'s `port: 8080`, the compose healthcheck's `localhost:8080`, and the `:5432` / `:8000` in the two `PS_*_URI` values are all container-internal and all correctly left alone. |
| Interaction with sibling #51 | ✅ Coherent, no ordering dependency beyond the README conflict. #51's `100_auth_users.sql` lands in the shared `supabase/seeders/sample_data/`, which `supabase db reset` loads for whichever project the local `config.toml` names — so after this PR its published-password credential lands in the `-e2e` stack on 55321. That is the right place for it, and #51's safety argument is unaffected in either direction. |
| `powersync/compose.yaml` network + internal port | ✅ The `external: true` network is renamed consistently in all four places it appears, and the failure mode if the network is absent is legible. The healthcheck correctly still targets the container-internal 8080; only the published host port needed to move, and only the published host port did. |
| `powersync/.env.example` | ✅ Round-1 Blocker 2 genuinely closed here. Both container hostnames renamed correctly, and the `:5432` / `:8000` suffixes rightly left alone — container-internal, not published. |
| `powersync/README.md` migration recipe | ✅ Every command actually runs. The "What this checkout is for" section is the strongest piece of writing in the PR: it states the limitation plainly, names `E2E_BACKEND_DIR` and its sibling-directory default explicitly, and does not oversell what the rename buys. Round-1 Nit 4 properly closed. |
| `docs/PowerSync.md` | ✅ No issues. One host-machine JWKS example bumped 54321 → 55321; the Docker-internal and Supabase Cloud variants directly above and below it are correctly untouched. |
| Both `global_search` READMEs | ✅ No issues. Correct derived container name, so the hand-apply recipes keep working. |

**One caveat on scope.** Docker was not started on this machine, so container and volume naming is verified *by construction* — from the CLI's documented `supabase_<service>_<project_id>` scheme, from Compose's project-name derivation, and from the fact that the pre-rename arrangement has worked since #37 — rather than by bringing a live stack up. Blocker 2's mechanism does not depend on a live run (a `container_name` is global by definition), but the specific failure text Docker emits could differ from the wording quoted.

</details>

<details>
<summary><b>Out of scope — flagged, not counted against this PR</b></summary>

- **Forcing `E2E_BACKEND_DIR` to a second checkout.** Lives in `construculator-app`. This PR makes a two-checkout setup *possible*; nothing makes it *happen*. No follow-up ticket is cited.
- **Self-healing `e2e_ensure_powersync_env()`.** Also `construculator-app`. Correctly declined in round 1 as out-of-repo.
- **The stray legacy `.github/database_tests.yml`** — a 267-byte fragment with no jobs. Pre-existing, unaffected by this PR, but worth deleting some day.

</details>

---

### Suggested next steps

- [ ] 🔴 **`construculator-app` `scripts/e2e/lib.sh`** — set `E2E_SUPABASE_API_PORT=55321`, `E2E_SUPABASE_DB_PORT=55322`, `E2E_MAILPIT_PORT=55324`, `E2E_POWERSYNC_PORT=9081` in app#576, then merge the two PRs together. Neither can land alone.
- [ ] 🔴 **`powersync/compose.yaml`** — add a top-level `name: construculator-backend-e2e-powersync` and remove the three `container_name` pins (L6, L16, L31), so two checkouts get separate PowerSync containers and separate bucket storage.
- [ ] 🟨 **`README.md:69-70`** — drop the *"or use a different `project_id` in your own checkout"* clause; it does not isolate anything and recreates the bug the PR is fixing.
- [ ] 🟨 **`construculator-app` `docs/Testing/CI-Scripts.md:304`** + the Codemagic `construculator_dev` environment group — move `SUPABASE_URL` from 54321 to 55321.
- [ ] 🟨 **`powersync/README.md:37`** — make the non-destructive `npx supabase stop --project-id construculator-backend` the copy-paste line and demote `--no-backup` to an opt-in comment.
- [ ] ⚫ **`powersync/README.md:46`** — drop the `cp .env.example .env`; stopping at `rm -f` lets `e2e_ensure_powersync_env()` regenerate the file with a fresh `PS_API_TOKEN`.
- [ ] ⚫ **`supabase/config.toml:88-89`** — shift the commented `smtp_port` / `pop3_port` to `55325` / `55326` for consistency with the block they sit in.
- [ ] ⚫ **`supabase/config.toml`** — consider a port block below 49152 (e.g. `443xx`); `553xx` is inside the OS ephemeral range and will occasionally fail to bind.
- [ ] ⚫ **Merge order with #51** — land #51 first, then drop this PR's README hunk for the deleted section; whichever lands second, re-read the *Local Services & Ports → Signing In* stretch for surviving 543xx references, because the conflict will not flag them.
- [ ] ⚫ **Open a tracked ticket** for the remaining structural gap (E2E harness targeting a dedicated second checkout) rather than leaving it as PR-thread prose. Your reply commits to "flagging to the team"; a ticket number in the README note would make the gap survive the merge.

powersync/compose.yaml
ayanasamuel8: Fixed in 850ea44 — added `name: construculator-backend-e2e-powersync` and dropped the three `container_name` pins, exactly as suggested. Verified live: brought up the stack and confirmed containers now come up as `construculator-backend-e2e-powersync-{mongo,powersync}-1` with a per-project `mongo_storage` volume, so two checkouts genuinely get separate PowerSync stacks now. Also updated the two `docker logs powersync` references in `powersync/README.md` to `docker compose logs powersync`, since the container no longer carries a fixed name — confirmed that command works against the renamed container. Resolving.

README.md
ayanasamuel8: Fixed in 850ea44 — dropped the "or use a different `project_id`" clause, you're right it doesn't isolate anything and would just point the destructive scripts at the renamed project. The note now says only the separate-clone remedy, matching what `powersync/README.md` already said. Resolving.

powersync/README.md
ayanasamuel8: Fixed in 850ea44 — inverted the default. The copy-paste line is now `npx supabase stop --project-id construculator-backend` (no `--no-backup`), with the destructive flag demoted to a follow-on comment for readers who've confirmed they don't need the data. Resolving.

ayanasamuel8:
Addressed the round-3 nit "silent merge-conflict risk with sibling PR #51 — README.md" now that #51 has merged.

Merged `origin/master` into this branch (62ac14d). As predicted, the only conflict was in `README.md`'s "Signing In as the Seeded Test User" section — this branch still had the old manual Studio-linking steps (with the port bumped to `24323`), while #51 replaced that content with the fixed `seeder@example.com` / `e2e-local-only-password` credentials.

Resolved in #51's favor per the reviewer's own recommendation, then re-checked the surrounding "Local Services & Ports → Signing In" stretch by hand rather than trusting the auto-merge:
- Full-tree grep for `5432[0-9]`, `8081`, `8083` post-merge: zero hits outside the already-reviewed explanatory comment/commented-out lines.
- `100_auth_users.sql` (from #51) seeds cleanly on top of this branch's `243xx` ports — ran `supabase db start` + `supabase test db` locally: 198/198 tests pass, including `seeded_auth_user_test.sql`.
- Ran the full stack (`supabase start`) and replicated the exact HTTP check the app repo's `E2E Environment Smoke` workflow runs: `POST /auth/v1/token?grant_type=password` for `seeder@example.com` → `200`, valid `access_token` returned.

The second round-3 nit (exporting the port block for `construculator-app` to source programmatically) is a forward-looking suggestion the reviewer explicitly scoped as "not a blocker for this PR" — leaving that as a follow-up rather than in-PR work.

